PR #200, opened 2014-03-27
fix windows build

m4r71n:
was broken with efff5a3daeb0571bb3fe8345b72d3f77b79d0066

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -35,7 +35,7 @@
 
 #ifndef COPYQ_TRANSLATION_PREFIX
 #   ifdef Q_OS_WIN
-#       define COPYQ_TRANSLATION_PREFIX QApplication::applicationDirPath() + "/translations"
+#       define COPYQ_TRANSLATION_PREFIX QCoreApplication::applicationDirPath() + "/translations"
 #   else
 #       define COPYQ_TRANSLATION_PREFIX ""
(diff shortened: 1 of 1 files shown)


src/app/app.cpp
```diff
@@ -35,6 +35,7 @@
 
 #ifndef COPYQ_TRANSLATION_PREFIX
 #   ifdef Q_OS_WIN
+#       include <QApplication>
 #       define COPYQ_TRANSLATION_PREFIX QApplication::applicationDirPath() + "/translations"
```
hluk: Can you rather change `QApplication` on this line to `QCoreApplication`?

m4r71n: I'll try it but not before evening

m4r71n: No unfortunately not. Btw. `QCoreApplication` is already included in [line 29](https://github.com/hluk/CopyQ/pull/200#discussion-diff-11055958L29)

hluk: I actually meant using `QCoreApplication::applicationDirPath()` on the line below so the header `QApplication` doesn't need to be included here.

m4r71n: Sorry that was a misunderstanding from me. I've updated the change.

hluk: No problem, thanks.

PR #307, opened 2015-01-03
Fix path to dmg file for travis deploy.

rygwdn:


rygwdn:
Looks like everything else about the travis deploy worked, but it didn't find any files to upload

PR #348, opened 2015-05-23
Added apt-get install command

arthurkarganyan:


hluk:
What about cmake package?

arthurkarganyan:
Does it install all this packeges?

hluk:
You need to install `cmake` explicitly since CopyQ is built using CMake.

You won't need `cmake` and `-dev` packages when CopyQ is installed from a package.

For reference [here is file for Debian package](https://build.opensuse.org/package/view_file/home:lukho:copyq/CopyQ/debian.control?expand=1). `Build-Depends` section lists packages needed to build and `Depends` lists packages needed to run CopyQ.

arthurkarganyan:
Thanks! I sorry I didn't know.

PR #496, opened 2016-04-26
Fix pasting with two monitors

m4r71n:
See comments from 07012d940415f7344f6682ef2e2545407a6d8a11


hluk:
@m4r71n OK, I didn't think testing this with multiple monitors is needed. Thanks :).

I need to check how the name changes with even more monitors -- maybe it'll be enough to test if the window class name doesn't start with `Shell_` and end with `TrayWnd`.

PR #568, opened 2016-11-05
[commandhelp.cpp] "arguments" is array

guraga:
Use square brackets ("[", "]") instead of round brackets ("(", ")"), because "arguments" object is array (not function).

Also need fix all translated files.


hluk:
Nice find!

PR #770, opened 2017-08-24
Fix cmake errors on eg. nixos

hedning:
When trying to package copyq for nix cmake couldn't find all qt modules, see
https://github.com/NixOS/nixpkgs/issues/26160 for a similar issue with kdenlive. Not well versed in cmake, but this seems like a reasonable fix. (love copyq btw :)


hluk:
Nice, thanks!

PR #777, opened 2017-09-03
Update readme.md: add windows choco command

duianto:
Add the chocolatey install command ~~, and the optional install parameters for:
excluding the desktop icon, not starting with the OS, and adding a shortcut.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,16 @@ For unlisted systems, please follow the instructions in
 
 ### Windows
 
-On Windows you can install [Chocolatey package](https://chocolatey.org/packages/copyq).
+On Windows you can install the [Chocolatey package](https://chocolatey.org/packages/copyq).
+
+```
+choco install copyq
+```
+
+#### Optional choco parameters
+- `/NoDesktopIcon` - Do not create a desktop icon
+- `/NoStartup` - Do not run CopyQ on OS startup
+- `/ToggleShortcut` - Specify a global shortcut to launch copyq, for example `/ToggleShortcut:ctrl+q`
 
 ### OS X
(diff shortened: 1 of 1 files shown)


duianto:
I removed the choco parameters, leaving only the `choco install copyq` instruction.

The parameters have several issues.

When installing with the choco command from a (suggested) administrator command prompt on a standard windows account. Then `copyq` only starts automatically when logging in, to the administrator account, and a quick test didn't seem to apply `/toggleshortcut:ctrl+q`. I'm just guessing but maybe the toggleshortcut is added to the copyq desktop icon, and if one chooses not to add the desktop icon then it has no effect.

The chocolatey copyq page: https://chocolatey.org/packages/copyq
only lists the three available package parameters:
>/NoDesktopIcon - Do not create desktop icon
/NoStartup - Do not run CopyQ on OS startup
/ToggleShortcut - Specify global shortcut to use to launch copyq, for example /ToggleShortcut:ctrl+q.

but I noticed on the visual studio code chocolatey page: https://chocolatey.org/packages/VisualStudioCode
that it also includes these two line:
>These parameters can be passed to the installer with the use of --params.
For example: --params '"/NoDesktopIcon /DontAddToPath"'.

And a quick test showed that the `--params '"parameters"'` command also is needed when installing copyq.

hluk:
The arguments after `--params` are perhaps just passed to the installer which is in this case Inno Setup ([list of arguments](http://www.jrsoftware.org/ishelp/index.php?topic=setupcmdline)).

But I think the install sections in README should be brief and easy to follow.

Thanks for PR!

PR #797, opened 2017-09-17
Log files location and access

orschiro:


docs/faq.rst
```diff
@@ -145,6 +145,8 @@ and set ``COPYQ_LOG_FILE`` to a file path for the log.
 You can copy current log file path to clipboard from Action dialog (F5 shortcut)
 by entering command ``copyq 'copy(info("log"))'``.
 
+Alternatively, press `F12` to directly access the log or find your log files `~/.local/share/copyq/copyq/copyq.log`.
```
hluk: The shortcut should be written as:
```
``F12``
```

Also, the given path is only for Linux (when using default XDG paths), it can be confusing for Windows and OS X users.

BTW, getting the path is described in the paragraph above (though it expects CopyQ to be running, which might not be the case, so it might be good idea to list the default paths for other platforms).

orschiro: I can add them. What are the default paths for the other platforms?

hluk: Actually, I have no idea, depends on lot of things. Maybe it's better to leave it out.

orschiro: Agree. How about just mentioning the useful F12 shortcut? :-)

hluk: Yup, that's the missing information. :)

orschiro:
@hluk have you seen my latest commit? Can we merge it like this?

hluk:
@orschiro Oh, completely missed the update (only noticed the comment before). Merged, thanks.

orschiro:
Thanks! :-)

PR #802, opened 2017-09-27
CMakeLists.txt: install appdata to /usr/share/metainfo

yangfl:
Ref: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html


hluk:
Thanks.

PR #857, opened 2018-01-12
Update build instructions to fix build error on macOS

ajorpheus:
Fixed the app name as it seems to be case sensitive.

This change fixes the following error:
$ make copyq.app
make: *** No rule to make target `copyq.app'.  Stop.


hluk:
Thanks. (Looks like it was tested before on a case-insensitive file system.)

ajorpheus:
No worries. Thanks for the awesomness that is CopyQ 👍

PR #877, opened 2018-02-18
Add RHEL/CentOS dependencies to Build from Source

eakst7:
Add the RHEL/CentOS yum command to install dependencies required to build to the Building from Source section.


hluk:
Thanks for the PR.

Did you check [copyq.spec for Fedora](https://src.fedoraproject.org/rpms/copyq/blob/master/f/copyq.spec)? Packages `libXtst-devel`, `libXfixes-devel` are good to have, not necessary but makes some features more stable.

eakst7:
Thanks, I updated the PR with the Xtst and Xfixes libs.

I also updated the docs in addition to the README.

hluk:
Nice, thanks again.

PR #1001, opened 2018-09-09
Improve consistency of scripting API page

PhiLhoSoft:


docs/scripting.rst
```diff
@@ -5,8 +5,8 @@ If you need to process items in some non-trivial way you can take
 advantage of the scripting interface the application provides. This is
 accessible on command line as ``copyq eval SCRIPT`` or
 ``copyq -e SCRIPT`` where ``SCRIPT`` is string containing commands
-written in Javascript-similar scripting language (Qt Script with is
-ECMAScript scripting language).
+written in JavaScript-similar scripting language (Qt Script with its
+ECMAScript scripting language, currently equivalent to ES5).
```
hluk: I think I meant "Qt Script is ECMAScript scripting language". It's based on [ECMA-262](http://doc.qt.io/qt-5/qtscript-index.html#language-overview).

PhiLhoSoft:
While I am at it, I would add a line indicating what is the meaning of the boolean value returned by copy(). Can you explain it here, please?

hluk:
> While I am at it, I would add a line indicating what is the meaning of the boolean value returned by copy(). Can you explain it here, please?

Oh, that return value is there only because backwards compatibility. Long time ago it returned false if the copy operation failed. Now it throws exception instead and it always returns true.

Maybe the `bool` in these cases (all `copy(...)` and `copySelection(...)` functions) should be removed from docs because it's confusing.

docs/scripting-api.rst
```diff
@@ -874,13 +882,21 @@ omitted.
...
+
+   Set current tray and window tag text.
+
+.. js:function:: String iconTagColor()
 
-   Get or set current tray and window tag text.
+   Get current tray and window tag color name.
 
-.. js:function:: iconTagColor([colorName])
+.. js:function:: iconTagColor()
```
hluk: Missing `colorName` argument.

PhiLhoSoft: Argh, I did the same error on iconColor(), and fixed it in time, but this one went unnoticed. Thanks.

hluk:
Thanks, the API documentation looks much more consistent.

PR #1003, opened 2018-09-11
Improve French translation

PhiLhoSoft:
... a lot! From simple missing (or extra) plural to opposite meaning (input vs. output!),
including improving consistency of translations, etc.

Notes:

There is a `<\p>` "tag" in the source text of translations.
For location filename="../src/ui/actiondialog.ui" line="112"

Is "Tab is full. Failed to remove any items." correct? Is it really "remove"?
For location filename="../src/gui/clipboardbrowser.cpp" line="792"


hluk:
> Is "Tab is full. Failed to remove any items." correct? Is it really "remove"?

This error pops up when adding an item into tab which reached item limit and no items can be removed from end (e.g. all are pinned).

Do you think I should rephrase it?

PhiLhoSoft:
I would say "don't worry, this message is very unlikely to appear, it isn't worth requesting updates from all translators for this one", at least until you make more important changes...

But perhaps this can be phrased as: "Tab is full. Cannot make room because all items are pinned." (or "unremovable") or something like that? Giving a hint on how to fix it.

Would need a setting with a low number of items per tab and / or lot of pinned items to get this one...

hluk:
Let's merge this now before there are any conflicts with Weblate translations.

I'll try update some of the messages after this is merged.

Thanks.

PR #1037, opened 2018-11-06
Use GNUInstallDirs to install arch-independent data

kapsh:
This allows better control on installation directories which helps on
distributions with multiarch layout. Executables and libraries go into
/usr/arch/{bin,lib} and architecture independent files will be installed in /usr/share.


hluk:
Nice, thanks.

PR #1053, opened 2018-12-03
Update install instructions for Arch Linux

maximbaz:
Hi there, I added `copyq` to the official Arch Linux repository, so that it's easier to install it. Thanks for this tool, keep up the good work 😉


hluk:
That's awesome! Thanks!

Issue #891, opened 2018-03-01
Add support in cmake script for creating app bundle on MacOS


meoow:
Qmake needs that big fat Xcode, cmake doesn't. I am not a developer, I just like compiling things by myself. So I don't really need Xcode which consumes a large portion of my storage. 

It would be nice to use cmake universally for cross-platform compilation. I managed to add a few lines in the CMakeLists files that can achieve the goal. It's based on v3.2.0.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f63b17e2..e62885d5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,6 +202,11 @@ if (UNIX AND NOT APPLE)
     add_definitions( -DCOPYQ_ICON_NAME="${ICON_NAME}" )
 endif()
 
+if (APPLE)
+    file(GLOB copyq_THEMES shared/themes/*.ini)
+    install(FILES ${copyq_THEMES} DESTINATION ${CMAKE_BINARY_DIR}/copyq.app/Contents/Resources/themes)
+endif()
+
 add_definitions( -DQT_NO_CAST_TO_ASCII )
 
 add_subdirectory(src)
@@ -210,3 +215,6 @@ if (WITH_PLUGINS)
     add_subdirectory(plugins)
 endif()
 
+if (APPLE)
+install(CODE "execute_process(COMMAND macdeployqt copyq.app WORKING_DIRECTORY \${CMAKE_BINARY_DIR})")
+endif()
diff --git a/plugins/CMakeLists.txt b/plugins/CMakeLists.txt
index 7c3caecb..645bcdda 100644
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -29,11 +29,19 @@ macro (copyq_add_plugin)
         qt4_add_resources(copyq_plugin_RCC ${copyq_plugin_${copyq_pkg}_RESOURCES})
     endif()
 
+    if (APPLE)
+        add_library(${copyq_pkg} SHARED
+            ${copyq_plugin_SOURCES}
+            ${copyq_plugin_FORMS_HEADERS}
+            ${copyq_plugin_RCC}
+            )
+    else()
     add_library(${copyq_pkg} MODULE
         ${copyq_plugin_SOURCES}
         ${copyq_plugin_FORMS_HEADERS}
         ${copyq_plugin_RCC}
         )
+    endif()
 
     set_target_properties(${copyq_pkg} PROPERTIES
[...]

Issue #590, opened 2016-11-27
The last several releases for macOS never work.


meoow:
```
$ ./copyq
dyld: Library not loaded: /usr/local/Cellar/qt5/5.7.0/lib/QtWidgets.framework/Versions/5/QtWidgets
  Referenced from: /Applications/CopyQ.app/Contents/Frameworks/QtSvg.framework/Versions/5/QtSvg
  Reason: image not found
Trace/BPT trap: 5
```

```
$ showdeps copyq
copyq:
	/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa
	@rpath/QtWidgets.framework/Versions/5/QtWidgets
	@rpath/QtNetwork.framework/Versions/5/QtNetwork
	@rpath/QtSvg.framework/Versions/5/QtSvg
	@rpath/QtXml.framework/Versions/5/QtXml
	@rpath/QtScript.framework/Versions/5/QtScript
	@rpath/QtMacExtras.framework/Versions/5/QtMacExtras
	@rpath/QtGui.framework/Versions/5/QtGui
	@rpath/QtCore.framework/Versions/5/QtCore
	/usr/lib/libc++.1.dylib
	/usr/lib/libSystem.B.dylib
	/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
	/System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation
	/usr/lib/libobjc.A.dylib
```

```
$ showdeps /Applications/CopyQ.app/Contents/Frameworks/QtSvg.framework/Versions/5/QtSvg
/Applications/CopyQ.app/Contents/Frameworks/QtSvg.framework/Versions/5/QtSvg:
	@executable_path/../Frameworks/QtSvg.framework/Versions/5/QtSvg
	/usr/local/Cellar/qt5/5.7.0/lib/QtWidgets.framework/Versions/5/QtWidgets
	/usr/local/Cellar/qt5/5.7.0/lib/QtGui.framework/Versions/5/QtGui
	/usr/local/Cellar/qt5/5.7.0/lib/QtCore.framework/Versions/5/QtCore
[...]

PR #1113, opened 2019-03-24
Build an App Bundle on macOS with CMake and CPack

briskycat:
Finishes the work started in #891
How to use:

1. `cmake --build . --target all --parallel`
2. `cpack`

You should get something like `CPack: - package: /.../copyq-3.7.3-76-g5f3a5138-Darwin.dmg generated.` at the end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 3.1)
 project(copyq)
+
+if (NOT APPLE)
+    set(COPYQ_EXECUTABLE_NAME ${PROJECT_NAME})
+else()
+    set(COPYQ_EXECUTABLE_NAME "CopyQ")
+endif()
 
 # C++11
 if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.1)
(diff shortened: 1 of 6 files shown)


briskycat:
Yes, I'll try.

briskycat:
Done. 

Not sure why you have these
```
root=$PWD

...

cd "$root"
```
commands in the scripts (they don't make any sense to me as current working directory is not propagated to shell's parent process) but I left them as they were anyway.

hluk:
Thanks, this helps a lot. I fixed the failing test (issue in fakevim plugin tests).

Can you also verify that the build binary doesn't use Qt libraries installed on system? I think following command should print Qt libs only from the app bundle (that's what the `macdeployqtfix` tool fixed before).

```bash
otool -L CopyQ.app/Contents/Frameworks/QtSvg.framework/Versions/5/QtSvg
```

For details see [this comment](https://github.com/hluk/CopyQ/issues/590#issuecomment-264120833).

Otherwise it looks good to merge. Thanks to this I'll no longer need to update and keep those obsolete *.pro files (I just need to figure out how to update translations).

briskycat:
Added otool invocation to `script-osx.sh`. Also added a step to uninstall local qt5 before the tests.

hluk:
Awesome! Thanks again.

PR #1116, opened 2019-03-29
Remove invalid doc for updateClipboardData(). 

akwala:
updateClipboardData() documented twice. Remove the invalid doc -- validity based on the function definition in /src/scriptable/scriptable.cpp.


hluk:
Thanks.

Issue #1163, opened 2019-06-12
can't start copyq: QLocalServer::listen: Name error


goldylucks:
**Describe the bug**
running `$ copyq` from terminal doesn't load the app and logs the following:

```bash
$ copyq
ERROR: Failed to create server "": QLocalServer::listen: Name error
Warning: CopyQ server is already running.
```

**To Reproduce**
I don't know how to reproduce on another machine

**Expected behavior**
Open copyq when typing `$ copyq` in terminal

**Screenshots**
N/A

**Version, OS and Environment**
```bash
$ copyq version
CopyQ Clipboard Manager v3.8.0
Qt: 5.9.5
Compiler: GCC
Arch: x86_64-little_endian-lp64
OS: Ubuntu 18.04.2 LTS
```

I'm not sure when this started happening.

some more outputs:

```bash
$ copyq show
ERROR: Cannot connect to server! Start CopyQ server first.

$ copyq tab 
ERROR: Cannot connect to server! Start CopyQ server first

$ copyq &  
[1] 11424
ERROR: Failed to create shared memory: QSharedMemory::create: already exists
Warning: CopyQ server is already running.
[1]  + 11424 done       copyq

$ sudo copyq
ERROR: Failed to create server "": QLocalServer::listen: Name error
Warning: CopyQ server is already running.
```

Any ideas?

I tried also `$ sudo apt purge copyq` and reinstalling it `$ sudo apt install copyq` with no success

hluk:
What's the output of following commands?

`copyq info`

`ls -ld ~/.config/copyq`

`ls -l ~/.config/copyq/.copyq*`

goldylucks:
thank you for the quick response @hluk !

```bash
$ copyq info
config: /home/goldy/.config/copyq/copyq.conf
desktop: /usr/share/applications/com.github.hluk.copyq.desktop
exe: /usr/bin/copyq
has-global-shortcuts: 1
has-mouse-selection: 1
icons: /usr/share/icons/hicolor/scalable/apps/copyq
log: /home/goldy/.local/share/copyq/copyq/copyq.log
platform: Linux/X11
plugins: /usr/lib/copyq/plugins
themes: /usr/share/copyq/themes
themes(custom): 
translations: /usr/share/copyq/translations
translations(custom): 

$ ls -ld ~/.config/copyq
drwxrwxr-x 2 goldy goldy 4096 Jun 13 11:02 /home/goldy/.config/copyq

[...]

PR #1164, opened 2019-06-13
add prompting for useful info in bug report

goldylucks:
I recently submitted a bug #1163  and was asked to provide the output of the commands this commit adds. Putting that in the bug report template will further streamline the process.


hluk:
Thanks, but those commands are useful only for specific issues, additionally, the `ls` commands could work only on Linux.

The description in your issue (#1163) is very good, the problem is that it's very unusual and it can be caused by a lot of things, so I'm just guessing what could be wrong.

goldylucks:
gotcha, wanted to streamline the process for you ;)

I'm closing this, if you change your mind just reopen it and tag me, or straight away merge it.

PR #1168, opened 2019-06-23
Spelling: MIME → media type, disable → turn off

comradekingu:
As per https://www.iana.org/assignments/media-types/media-types.xhtml


hluk:
Thanks.

Issue #1178, opened 2019-07-25
Triggering command based on primary selection only


ngirard:
On Linux, I wish an action to be triggered based on the contents of the PRIMARY selection only, not the CLIPBOARD selection.

Is it currently possible ?

hluk:
You can check `application/x-copyq-clipboard-mode` format in automatic commands.

E.g. if you set input format of a command it would be only executed on X11 selection change:

```ini
[Command]
Automatic=true
Command="
    copyq:
    popup(input())"
Input=application/x-copyq-clipboard-mode
Name=Executed only on X11 selection change
```

Otherwise you can check it in command:

```ini
[Command]
Automatic=true
Command="
    copyq:
    if (str(data(mimeClipboardMode)) == 'selection')
      popup('selection changed')
    else
      popup('clipboard changed')"
Name=Show clipboard/selection change
```

ngirard:
Thanks very much for your feedback !

I just created a PR that adds this to the FAQ.

PR #1180, opened 2019-07-27
FAQ: added "How to trigger a command based on primary selection only?"

ngirard:
As per #1178 .

Code is yours, but phrasing is mine... you might want to edit it.


hluk:
Thanks.

PR #1205, opened 2019-08-28
typo replaced "applicaton" with "application"

leela52452:


hluk:
Thanks! Though fixing it this way won't work. This needs to be fixed in the source code (`*.cpp` files). The translation files (`*.ts`) are generated from the source.

I can fix that and regenerate the translation if that's OK with you.

hluk:
To be more clear: The texts need to be changed in `*.cpp` files; `*.ts` files are generated semi-automatically (by a utility).

leela52452:
i did not see your message
does deleting my repo will remove this mess

hluk:
I just checked that those translation strings are marked as "vanished", i.e. already removed from UI. (It's better to keep old strings around so some translation apps can make nicer suggestions for similar texts.)

Since this is not needed, I'm closing this.

PR #1261, opened 2019-11-20
Add emacsclient example editor command for Windows

pcrama:
This example (for Windows) assumes
- that copyq shares the same HOME environment variable as your normal emacs installation
- that emacsclientw.exe and runemacs.exe are on the PATH (in copyq environment)
- that the emacs server is setup

(tested with emacs & copyq installed via scoop and a starter script setting HOME in the environment before starting copyq)


hluk:
Merged, thanks.

PR #1322, opened 2020-02-27
Update basic-usage.rst

gregfenton:
To people new to OSX, the "tray" is assumed to be Launchpad (which it isn't).  Clicking the icon from Launchpad and/or Finder does not open any windows or provide any feedback.  Unless you catch that there's a new tiny icon in the tray, one assumes that the app just isn't working.  I wonder how many people have been confused by this and abandoned the app . . .


hluk:
Thanks, merged.

PR #1327, opened 2020-03-08
Bind x to Delete in Vi style navigation mode

stepnem:
It's so vexing having to reach off the home row :-)

x seems the natural choice.


hluk:
Thanks!

PR #1388, opened 2020-05-21
Update copyq_zh_CN.ts

v3470:
Update translations: fix typo and use proper classifier


hluk:
Thanks!

Ideally, if you could use web UI to translate (https://hosted.weblate.org/projects/copyq/master/); merging pull requests that change translation could lead to conflicts.

PR #1460, opened 2020-08-19
Add bash completion script

LiteracyFanatic:
This is a work in progress which could definitely use a refactoring, but I thought I'd submit it for feedback since I have it in a functional state. It would be great if other people could test it on their systems to make sure it works properly. Just run `source utils/copyq-completion` from a bash shell. The type `copyq ` and press `tab`.

There are comments explaining why some of the commands don't have completions, but I'll reiterate my questions here for clarity.

1. Do we want any sort of completion for mime types? Obviously there are a great many of them, but perhaps there is some useful subset that would make sense to list. For `clipboard`, `selection`, and `read` in particular I wonder if there is a way to query the available mime types for the current data. That would be nice for discoverability.
2. It would be possible to generate row numbers for the commands which take them, but since there would likely be a bunch of them this probably isn't a great idea.
3. Would the acceptable values of the `PROGRAM` argument to `action` be any executable on the `PATH`?
4. Is there an easy way to generate a list of allowed values for the config options that are finite sets i.e. booleans, enums, magic strings? It'd be nice to have, but I wouldn't want to hardcode it.
5. What are sessions? How would I get a list of available sessions?
6. Which commands are allowed as arguments to `tab`?
7. Is `utils/` the correct place for this script to live, or does it belong elsewhere?


LiteracyFanatic:
Apologies that it took me so long to get back to this. I've push changes adding completion for MIME types and sessions and checks to make sure the server is running when necessary.

I'm not familiar with CMake, but I took a stab at editing `CMakeLists.txt` anyway. I opted to install the completion script to `/usr/share/bash-completion/completions/` since many packages on my system seem to install there and let the `bash-completion` package handle loading them. I like this approach since it should just work for a lot of users without any additional setup. Alternatively we could put it under `/usr/share/copyq` and let users/packagers choose what to do with it if they want.

The script doesn't currently address that fact that you can continue listing commands after the first one. It should be possible with some sort of recursive function, but it would also have to do bookkeeping on the variable number of arguments for some commands and I think it would get really messy for only marginal returns.

For similar reasons I have only implemented completion for the commands covered in the man page. I can add more from the API page if you would like, but I think completions for ~160 commands would be a little overwhelming to present to the user. Plus the default settings for bash will display a prompt for long completions that the user would have to answer every time. `Display all N possibilities? (y or n)`

shared/copyq-completion
```diff
@@ -0,0 +1,106 @@
...
+
+commands=(
+    show hide toggle menu exit disable enable clipboard selection paste copy
+    count select next previous add insert remove edit separator read write
+    action popup tab removetab renametab exporttab importtab config eval session
+    help
+)
+
+list_sessions() {
+    ss -l -x | sed -n '/\/\.copyq[^_]*_s /{s/.*\/\.copyq-\?\([^_]*\)_s .*/\1/;p}'
```
hluk, line 11: I know I suggested this, but I think I missed the fact that CopyQ session name can contain `_`.

Here is fixed and simplified version:

```
ss -l -x | sed -nE 's#.*/\.copyq-?(\S*)_s .*#\1#p'
```

hluk (approved):
Nice! I think this is OK to be merged as is. I can fix the `sed` command later - I need to fix the tests for pull requests anyway (started using GitHub Actions just recently).

Thanks a lot!

PR #1499, opened 2020-10-23
Windows: Show window where cursor is

aibaron:
Is it possible to have the window be created where the cursor (either mouse or text cursor) is? Ditto does this well but CopyQ has many more features that I like.


hluk:
I'm confused. Why is this a pull request for a commit that is already merged in master?

hluk:
Did you only wanted to create a new issue? ... https://github.com/hluk/CopyQ/issues/new/choose

Pull requests are for submitting source code fixes/patches.

aibaron:
Thank you!

PR #1547, opened 2020-12-25
Fix installation instructions in documents

kawarimidoll:
Fix error when installing using latest Homebrew in macOS. Just fix `README.md` and `docs/installation.rst`.

screenshot:

Calling brew cask install is disabled


hluk:
Thanks for the documentation fix!

PR #1548, opened 2020-12-27
Namespace bash completions

LiteracyFanatic:
I was writing another completion script for one of my own projects today and I realized that this will bug out if another script uses the same function names because it's being sourced. This just adds a unique prefix to make sure that doesn't happen.

### DIFF
--- a/shared/copyq-completion
+++ b/shared/copyq-completion
@@ -1,29 +1,29 @@
 # shellcheck shell=bash
 
-commands=(
+_copyq_commands=(
     show hide toggle menu exit disable enable clipboard selection paste copy
     count select next previous add insert remove edit separator read write
     action popup tab removetab renametab exporttab importtab config eval session
     help
 )
 
-list_sessions() {
+_copyq_list_sessions() {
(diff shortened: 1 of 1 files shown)


shared/copyq-completion
```diff
@@ -1,29 +1,29 @@
...
     show hide toggle menu exit disable enable clipboard selection paste copy
     count select next previous add insert remove edit separator read write
     action popup tab removetab renametab exporttab importtab config eval session
     help
 )
 
-list_sessions() {
-    ss -l -x | sed -nE 's#.*/\.copyq-?(\S*)_s .*#\1#p'
+_copyq_list_sessions() {
+    ss -l -x | sed -n '/\/\.copyq[^_]*_s /{s/.*\/\.copyq-\?\([^_]*\)_s .*/\1/;p}'
```
hluk: Looks like this has been changed back accidentally.

LiteracyFanatic: Woops. Fixed.

hluk:
Thanks!

PR #1566, opened 2021-01-23
platform/x11: correct autostart boolean value type

hexchain:
According to the [desktop-entry-spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s04.html), a boolean value must be either `true` or `false` (in lowercase).


hluk:
Thanks.

PR #1607, opened 2021-03-18
Fix invalid return value if the temp file cannot be opened

juergenhoetzel:


hluk (approved):
Whoa, oops. Nice find! Thanks for fixing it.

PR #1625, opened 2021-04-12
Minor update: PEP8 formating and f strings

SUPERustam:
Very minor update


hluk:
Thanks but there is no need to keep strict formatting in these support scripts. I also prefer some of the current code.

SUPERustam:
Ok)

Issue #1628, opened 2021-04-13
Preferences menuitem leads to the Configuration dialog box


Gitoffthelawn:
**Describe the bug**
The `Preferences` menuitem leads to the `Configuration` dialog box in the English i18n.

I'll see if I can fix it right now and submit a pull request.

Gitoffthelawn:
Fixed via PR 1629: https://github.com/hluk/CopyQ/pull/1629

PR #1629, opened 2021-04-13
Fix for https://github.com/hluk/CopyQ/issues/1628

Gitoffthelawn:


hluk:
Ah, true, this is better and less confusing. Thanks!

Gitoffthelawn:
You're welcome!  And as always, thank you for CopyQ!

PR #1650, opened 2021-05-01
Fix installing bash completion script

LiteracyFanatic:
CMake is currently putting the wrong file in the bash completion directory. This fixes that.


hluk:
Thanks.

Issue #1395, opened 2020-05-29
Paste to current window not working (Win10)


deuntje:
**Describe the bug**
I'm configuring a Windows 10 (Citrix) VDI and try to get copyq working. I have this problem that my pastes from copyq are not being pasted in the current window (tried multiple applications: Notepad++, notepad.exe, . My selected item is copied to the clipboard, so a CTRL+V or Shift+Insert will paste this, but this will take an extra step which is not needed. 

I've tried many approached:
- install latest copyq (v3.11.1) and set it up from scratch
- install latest copyq (v3.11.1) and import settings from a working copyq version of my desktop-pc
- install latest copyq (v3.11.1) and copy %APPDATA%\copyq dir from working copyq version of my desktop-pc
- install older copyq v3.6.1 and set it up from scratch
- tried the paste_with_ctrl_v_windows=".*" option in General section of copyq.ini

**To Reproduce**
in all approaches I use Meta+Insert as global shortcut to open copyq and with Preference > History > ' After item is activated...' options enabled.

1. Go to notepad
2. META+Insert to open copyq
3. pick an item
4. Hit Enter

My selected item is not pasted in notepad, it is copied to the clipboard, so a CTRL+V or Shift+Insert will paste this (the focus is back to the notepad application), but this will take an extra step which is not needed. 

**Expected behavior**
Paste the selected item in notepad after the 'Enter'

**Version, OS and Environment**
 - Application Version: 3.11.1
 - OS: Windows 10 Enterprise 1903
 - Citrix VDI

hluk:
Can you see if there are any warnings or errors in logs (F12 from main window)? My guess is that CopyQ is either failing to get the correct focused window or send the Shift+Insert shortcut.

deuntje:
In the copyq log there are no logs. 

After the return/paste, copyq returns to the current window (Notepad++) and seems to successfully get the focus. When I manually Shift+Insert after that the item I selected got pasted. Seems that the copying towards clipboard succeeds, but their still is a focus/shift+insert issue.

Tried another program (SecureCRT), this program gives the error: "Could not open clipboard: Access is denied.". Also here, focus seems ok, after manually shift+insert the item will be pasted. 

While updating this issue I found this clipboard debug tool: https://pretentiousname.com/miscsoft/#ClipboardOwnerDebug	and find out what was causing the issue. 

```
C:\tools\ClipboardOwnerDebug_1001>ClipboardOwnerDebug.exe 100
[...]

PR #1686, opened 2021-06-21
Provide a fallback method for pasting in macOS (#1395)

r-darwish:
If the window isn't available post the shortcut directory to the PID


r-darwish:
@hluk can you please review?

hluk:
@r-darwish Nice! Thanks for the fix!

PR #1689, opened 2021-06-22
Fix handling of git GPG in Windows

r-darwish:
If CopyQ detect's the git bash gpg in path it will use it. However, this version of GPG expects unix paths instead of Windows. Trying to run it with Windows native paths cause it to hang for a lot of time, consuming 100% CPU.

This patch tried to detect whether the given version of GPG in Windows is a unix one, and if such, convert the key paths to unix

### DIFF
--- a/plugins/itemencrypted/itemencrypted.cpp
+++ b/plugins/itemencrypted/itemencrypted.cpp
@@ -54,18 +54,6 @@ const QLatin1String dataFileHeaderV2("CopyQ_encrypted_tab v2");
 
 const int maxItemCount = 10000;
 
-struct KeyPairPaths {
-    KeyPairPaths()
-    {
-        const QString path = getConfigurationFilePath("");
-        sec = QDir::toNativeSeparators(path + ".sec");
-        pub = QDir::toNativeSeparators(path + ".pub");
-    }
-
-    QString sec;
-    QString pub;
-};
-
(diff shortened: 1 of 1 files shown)


plugins/itemencrypted/itemencrypted.cpp
```diff
@@ -132,6 +132,25 @@ const QString &gpgExecutable()
...
+    KeyPairPaths()
+    {
+        const QString path = getConfigurationFilePath("");
+        sec = QDir::toNativeSeparators(path + ".sec");
+        pub = QDir::toNativeSeparators(path + ".pub");
+
+#ifdef Q_OS_WIN
+        if (checkUnixGpg(gpgExecutable())) {
+            pub.replace('\\', '/').replace(":", "").insert(0, '/');
+            sec.replace('\\', '/').replace(":", "").insert(0, '/');
```
hluk: How about using `QDir::fromNativeSeparators` instead of replacing the chars?

```suggestion
            pub = QDir::fromNativeSeparators(pub).replace(":", "").insert(0, '/');
            sec = QDir::fromNativeSeparators(sec).replace(":", "").insert(0, '/');
```

Or even better skip the previous `QDir::toNativeSeparators()` calls in this case.

r-darwish: Seems to work. Thanks

hluk:
Thanks!

PR #1697, opened 2021-07-06
Update Dependencies list for RHEL family

jacob-pro:
On Oracle Linux 8 (and presumably will for other RHEL family / Yum OS):

```
-- Could NOT find KF5Notifications: found neither KF5NotificationsConfig.cmake nor kf5notifications-config.cmake
```


hluk:
Thanks!

PR #1759, opened 2021-09-17
Added example to paste copied text as title case

viharm:
Reference: https://stackoverflow.com/a/196991


hluk:
Looks good, thanks. There is more advanced version in commands repo: https://github.com/hluk/copyq-commands/blob/master/Global/to-title-case.ini

PR #1769, opened 2021-10-01
Update Inno Setup script

bovirus:
Add info about

VersionInfoDescription
VersionInfoProductName
VersionInfoVersion
UninstallDisplayName
UninstallDisplayIcon
AppCopyright=

### DIFF
--- a/shared/copyq.iss
+++ b/shared/copyq.iss
@@ -26,6 +26,13 @@ SetupIconFile={#Source}\src\images\icon.ico
 WizardImageFile=logo.bmp
 WizardSmallImageFile=logo-small.bmp
 
+VersionInfoDescription=CopyQ installer
+VersionInfoProductName=CopyQ {#AppVersion}
+VersionInfoVersion={#AppVersion}
+UninstallDisplayName=CopyQ {#AppVersion}
+UninstallDisplayIcon={#Root}\copyq.exe
+AppCopyright=(c) Lukas Holecek
+
 [Languages]
 Name: en; MessagesFile: "compiler:Default.isl"
 Name: cz; MessagesFile: "compiler:Languages\Czech.isl"
@@ -78,8 +85,8 @@ it.PluginNotes=Note
 it.PluginEncrypted=Criptazione
 it.PluginFakeVim=Editor FakeVim
 it.PluginSynchronize=Sincronizza elementi nel disco
-it.PluginTags=Tag elementi
-
+it.PluginTags=Etichette elementi
+it.PluginPinned=Elementi bloccati
 
 [Types]
 Name: "full"; Description: "{code:GetFullInstallation}"


shared/copyq.iss
```diff
@@ -26,6 +26,13 @@ SetupIconFile={#Source}\src\images\icon.ico
 WizardImageFile=logo.bmp
 WizardSmallImageFile=logo-small.bmp
 
+VersionInfoDescription=CopyQ installer
+VersionInfoProductName=CopyQ {#AppVersion}
+VersionInfoVersion={#AppVersion}
```
hluk, line 31: This causes [the build to fail](https://ci.appveyor.com/project/hluk/copyq/builds/40986200/job/itsno61xayn6b9xe):

```
Error on line 31 in C:\projects\copyq\Shared\copyq.iss: Value of [Setup] section directive "VersionInfoVersion" is invalid.
```

Looks like the version should be strictly in format `w.x.y.z` (with missing numbers replaced by zeroes): https://jrsoftware.org/ishelp/index.php?topic=setup_versioninfoversion

I believe the `AppVersion` value is `git describe --tags --always HEAD` - in case of a dev build it would be something like `v5.0.0-13-g5fc45fd0` and in case of a release `v5.0.0`.

bovirus: @hluk 

Probably the error could be why "AppVersion" value (that I didn't touch and was already present in original script) should be a full numeric value and you are using "vx.x.x-yyyyy".

For this reason I suggest to use full numeric value like 5.0.0-xxxxxxx.

The value of Appversion in Inno Setup should be always full numeric value without alphabetic characters.

To fxi this sissue you canm comment the string

;VersionInfoProductName=CopyQ {#AppVersion}
;VersionInfoVersion={#AppVersion}

And ths ugegstion is to use base numeri version (ex. 5.0) and a build number (ex. 1000) and in this case you will have a full version like 5.0.1000.

New revision change only build number.

Thanks.

hluk: Can you transform the AppVersion to correct format for VersionInfoVersion in the Inno Setup script? E.g. `v5.0.0-13-g5fc45fd0` would become `5.0.0.13` and `v5.0.0` would become `5.0.0`.

I don't want to change the version in the app itself because knowing the exact dev build commit is useful.

bovirus: I will check.
For this moment I suggest to insert comemt to the line

;VersionInfoVersion={#AppVersion}

hluk: Let me drop the "v" version prefix globally as you suggested: #1771

bovirus: Probably remove the "v" abnd use the full numeric value (ex. 5.0.0) is the best way.

hluk: Can you rebase your change on new master? The problem should be fixed there (`AppVersion` is in correct format).

bovirus: Ok

PR #1796, opened 2021-10-30
Windows: Fix installing in 64bit path

### LINKED ISSUE #1792: Installs in the 32 bit 'Program Files (x86)' directory
'Windows: Builds are now 64bit' from 'https://github.com/hluk/CopyQ/releases/tag/v4.0.0' but installs to 'C:\Program Files (x86)\CopyQ' by default. Not a major issue but it is a bit confusing.

hluk:
The default install location for the 64bit app should be
"C:\Program Files".

Fixes #1792

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -204,11 +204,11 @@ On Windows (in PowerShell):
 
 .. code-block:: powershell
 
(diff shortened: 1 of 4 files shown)


kurtmckee:
@hluk I'd be happy to test this if you'd like. Do you have an development build available?

PR #1802, opened 2021-11-03
Add option to disable restoring window geometry

hluk:
Restoring window geometry can be disabled with following command:

    copyq config restore_geometry false

Application restart is needed after changing the option.


blueyed:
@hluk 
Thank you, that works for me to work around the placement issue(s).

PR #1890, opened 2022-01-25
docs: Update macOS building from source instructions

axonasif:
I was trying to follow the macOS build from source instructions but wasn't able to compile. Thankfully we can follow pretty much the same steps as in the github actions method :)

### DIFF
--- a/docs/build-source-code.rst
+++ b/docs/build-source-code.rst
@@ -124,7 +124,12 @@ On OS X, required Qt 5 libraries and utilities can be easily installed with `Hom
 
 ::
 
-    brew install qt5
+    cd CopyQ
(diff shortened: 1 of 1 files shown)


hluk:
Looks good. Thanks.

BTW, I recently found a tool to trigger the GitHub Actions locally - I have not tried it yet, but it could help to make a clean build with a simple command: https://github.com/nektos/act

axonasif:
> Looks good. Thanks.
> 
> BTW, I recently found a tool to trigger the GitHub Actions locally - I have not tried it yet, but it could help to make a clean build with a simple command: https://github.com/nektos/act

Wow, that's actually cool 👀

Issue #1846, opened 2021-12-12
Fix for "Change Copied Text to Title Case"


sohail-bakhtiari:
On the [Command Examples](https://copyq.readthedocs.io/en/latest/command-examples.html) page, the example for changing text to title-case does not work. It does nothing.

I believe there is some problem with `input()` in this line:
`copy(toTitleCase(str(input())))`

To make it work I used the previous example titled "Change Upper/Lower Case of Selected Text" to copy the selected text and paste the modified string instead:

```
  copyq:
    function toTitleCase(str) {
      return str.replace(
        /\w+/g,
        function(txt) {
          return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
        }
      );
    }

    if (!copy())
      abort()

    var text = str(clipboard())

    copy(toTitleCase(text))
    paste()
```

I also changed the regex from `/\w\S*/g` to `/\w+/g` to better process multipart words (words with dash or slash)

I'm not sure if it is OK to post this here, sorry if it is out of context.

hluk:
Can you create pull request? The documentation page source code is here: https://github.com/hluk/CopyQ/blob/master/docs/command-examples.rst

Kmfernan5:
This title case command is not working

https://github.com/hluk/CopyQ/blob/master/docs/command-examples.rst

Note [2022-04-25 19:09:27.250] <Server-15816>: Exception in command "Title Case": ScriptError: SyntaxError: Unterminated regular expression literal
Note [2022-04-25 19:09:27.250] <Server-15816>:    
Note [2022-04-25 19:09:27.250] <Server-15816>:    --- backtrace ---
Note [2022-04-25 19:09:27.250] <Server-15816>:    @eval code:4
Note [2022-04-25 19:09:27.250] <Server-15816>:    eval:function toTitleCase(str)...
Note [2022-04-25 19:09:27.250] <Server-15816>:    eval
Note [2022-04-25 19:09:27.250] <Server-15816>:    --- end backtrace ---
Note [2022-04-25 19:10:45.403] <Server-15816>: Exception in command "Title Case": ScriptError: SyntaxError: Unterminated regular expression literal
Note [2022-04-25 19:10:45.403] <Server-15816>:    
Note [2022-04-25 19:10:45.403] <Server-15816>:    --- backtrace ---
Note [2022-04-25 19:10:45.403] <Server-15816>:    @eval code:4
[...]

PR #2028, opened 2022-06-28
Correction of example snippet for changing text to title case

viharm:
This change makes a correction to the code snippet which changes the text to title case. The previous version had a syntax error with incorrectly escaped backslashes (`\`)

Related to #1846.


hluk:
Thanks for the PR.

The code block must be indented (4 spaces) so it shows correctly when rendered as rst. Preview is here: https://github.com/hluk/CopyQ/blob/9bef6d58004f4772aeb55f287c65353cdc898eb8/docs/command-examples.rst

viharm:
> Thanks for the PR.
> 
> The code block must be indented (4 spaces) so it shows correctly when rendered as rst. Preview is here: https://github.com/hluk/CopyQ/blob/9bef6d58004f4772aeb55f287c65353cdc898eb8/docs/command-examples.rst

I'm sorry about the schoolboy error. I've fixed it.

Issue #1903, opened 2022-02-08
Plugin "itemtags" is not installed on MacOS M1 variant.


axonasif:
**Describe the bug**
I'm unable to assign tags to clipboard items, maybe the tags plugin is missing in my macOS build?

**To Reproduce**
1. Compile copyq from source by following the instructions in the copyq docs.
2. Install and use copyq
3. Try to assign a tag on an item

**Expected behavior**
The itemtags plugin should be there and work properly

**Screenshots**
None

**Version, OS and Environment**
(Get details from `copyq version` command if possible.)
 - Application Version: 6.0.1-16-g963f64ef
 - OS: MacOS 12.2

axonasif:
Oh sorry @hluk 
I had the logs copied for forgot to paste 😅 

Here it goes:
```
Note [2022-02-08 12:30:44.538] <Server-15443>: No plugins loaded
Warning [2022-02-08 12:30:44.542] <Server-15443>: [kf.windowsystem] QtWarning: Could not find any platform plugin
Warning [2022-02-08 12:30:49.045] <Server-15443>: [qt.qpa.fonts] QtWarning: Populating font family aliases took 152 ms. Replace uses of missing font family "Monospace" with one that exists to avoid this cost. 
Note [2022-02-08 12:32:05.636] <Server-15443>: Exception in command "Add a Tag": ScriptError: Plugin "itemtags" is not installed
Note [2022-02-08 12:32:05.636] <Server-15443>:    
Note [2022-02-08 12:32:05.636] <Server-15443>:    --- backtrace ---
Note [2022-02-08 12:32:05.636] <Server-15443>:    get@:1
Note [2022-02-08 12:32:05.636] <Server-15443>:    %entry@eval code:1
Note [2022-02-08 12:32:05.636] <Server-15443>:    eval: plugins.itemtags.tag()
Note [2022-02-08 12:32:05.636] <Server-15443>:    eval
Note [2022-02-08 12:32:05.636] <Server-15443>:    --- end backtrace ---
Note [2022-02-08 12:32:05.638] <Server-15443>: Command “Add a Tag”
Note [2022-02-08 12:32:05.638] <Server-15443>:    Exit code: 4
Note [2022-02-08 12:32:05.638] <Server-15443>:    1. copyq:  plugins.itemtags.tag()
Note [2022-02-08 12:32:11.719] <Server-15443>: Exception in command "Add a Tag": ScriptError: Plugin "itemtags" is not installed
[...]

PR #2047, opened 2022-07-28
Update build-source-code.rst doc for #1903

tony-wallace:
I ran into the same problem as bug #1903, and the workaround proposed by @axonasif worked for me. 
This change simply tries to describe that workaround in the docs.


hluk:
Cool, thanks for sharing and updating the docs!

Issue #2038, opened 2022-07-09
"Show main window under mouse cursor" Dual Screen not going on 2nd monitor


rbooth84:
**Describe the bug**
When "Show main window under mouse cursor" is enabled when pressing the shortcut key the window will only popup on the last screen that it appeared on at the edge of the other screen. Also when "Close When Unfocused" is enable it will only close when focus is changed to another app that is on the same screen

**Steps to reproduce the behavior**
1. Enable "Show main window under mouse cursor"
2. Enable Close When Unfocused
3. Press the shortcut key on different screens.
4. When the window is displayed click on another app on the opposite screen as the window

**Expected behavior**
Should show at the mouse cursor regardless of what screen the cursor is on

**Version, OS and Environment**
 - Application Version: 6.2.0
 - OS: XeroLinux, Fedora KDE Spin
 - Desktop environment: X11 and KDE

**Additional context**
When using Zorin OS 16.1 which is based on Ubuntu 20.04 the issue does not exist.

presianbg:
Hi,

Same behavior in:
- Application Version: 6.2.0 , 6.1.0
- OS: Solus
- DE: X11 and Budgie

Cheers!

thennothinghappened:
I'm using a shortcut for "Show the Tray Menu" and it seems to have really weird inconsistent behaviour, sometimes it only goes on monitor 1, and other times only on monitor 2.
```
CopyQ Clipboard Manager 6.2.0
Qt: 5.15.5
KNotifications: 5.97.0
Compiler: GCC
Arch: x86_64-little_endian-lp64
OS: Arch Linux
```

PR #2096, opened 2022-09-21
Fix tray window popup on multi-monitor

thennothinghappened:
heya, me again!
I went through some of the code for the tray popup since it has an issue where it only sticks to the screen it was first shown on. I fixed that (`screen.cpp`) as it was reusing the previous screen, so it would only ever use the first one it used.
After that I did some more debugging and it seems that using `window->setPosition` seems to grab the previous position, so you have to show the menu twice in the same spot for it to get the correct coordinates. I removed that part of the code, and the menu now calculates the correct position for the usable area rather than sometimes using the one for the other monitor.

I've tested this on my PC, let me know if there's any issues on other setups. Thanks!

### DIFF
--- a/src/common/display.cpp
+++ b/src/common/display.cpp
@@ -56,11 +56,8 @@ int smallIconSize()
 
 QPoint toScreen(QPoint pos, QWidget *w)
 {
-    QWindow *window = w->windowHandle();
-    if (window)
-        window->setPosition(pos);
-    else
-        w->move(pos);
+	QWindow *window = w->windowHandle();
+    w->move(pos);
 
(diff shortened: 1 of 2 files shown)


thennothinghappened:
Confirmed this fixes #2038

hluk:
Thanks for the fix!

Does this work also on screens with different scaling/DPI?

thennothinghappened:
> Thanks for the fix!
> 
> Does this work also on screens with different scaling/DPI?

Yep! I tried differences with the resolution and DPI of both monitors and it works there too

hluk:
Nice, thanks again for the fix! :+1:

PR #2237, opened 2023-01-19
Drop settings backup handling and migrating commands

hluk:


hluk:
I'm going to leave this open for a while and keep using the patched app to see if it works properly.

Long time ago, without backing up the settings, Qt framework had problems updating the ini file (I'm guessing when modified from multiple processes or threads) leaving the file empty or otherwise corrupted.

jmlucjav:
Hi, 

I just tested the portable distrib from here https://ci.appveyor.com/project/hluk/copyq/builds/45985448/artifacts I overwrite my normally installed files (with chocolaty) with them, but I still see the same behavious, hardlinks are converted to normal files. Am I missing something?

thanks

hluk:
> I just tested the portable distrib from here https://ci.appveyor.com/project/hluk/copyq/builds/45985448/artifacts I overwrite my normally installed files (with chocolaty) with them, but I still see the same behavious, hardlinks are converted to normal files. Am I missing something?

Ah, this is probably caused by how Qt framework handles safe atomic file writes, see [QSaveFile details](https://doc.qt.io/qt-5/qsavefile.html#details):

> While writing, the contents will be written to a temporary file, and if no error happened, commit() will move it to the final file. This ensures that no data at the final file is lost in case an error happens while writing, and no partially-written file is ever present at the final location. Always use QSaveFile when saving entire documents to disk.

Probably nothing I can do about that.

Instead of linking the config file it would be possible to copy the file on app exit and restore on start but I don't think it would be very safe.

jmlucjav:
https://bugreports.qt.io/browse/QTCREATORBUG-19651 

I have never encountered an app that behaved like this, Qt seems like a big project, quite surprising they handle it so poorly (imho)...

thanks for the help!

jmlucjav:
By the way, this might be still fixed by hardlinking %AppData%\Roaming\copyq (for Win for instance) to my directory of choice, and then the files could be still handled as Qt wants, this would work. 

But then this directory maybe would have to be split between the config ini files (this is the one to be hardlinked), and another one where other files such as copyq_tab_JmNsaXBib2FyZA==.dat etc live, as those should be local to the installation, I think?

Not sure if you want to embark into this changes, anyway putting them here for reference.

hluk:
Merged since I haven't encountered any issue using this.

> But then this directory maybe would have to be split between the config ini files (this is the one to be hardlinked), and another one where other files such as copyq_tab_JmNsaXBib2FyZA==.dat etc live, as those should be local to the installation, I think?

The data files should go to a different directory (there is usually specific dir in system for this). I haven't found the time to do this.

PR #2238, opened 2023-01-19
Update installer script

bovirus:
@hluk 

More use of varables instead fixed text (ex. "CopyQ")
Use of Modern style
Disable "Usepreviouslanguage" setting to enable if new language added in new version will be detected.
Separated setting group with newline for better clearance

Note: for Copyright message could be ok for you
- to remove "(c)" not necessary because Copyright is field name
- to add at the end a starting copyright year/ending copyright year=current year? 
Which is for you your starting copyright year?

### DIFF
--- a/shared/copyq.iss
+++ b/shared/copyq.iss
@@ -1,25 +1,42 @@
-﻿; 1. Open this file with Inno Setup with Unicode support and preprocessor.
+; 1. Open this file with Inno Setup with Unicode support and preprocessor.
 ; 2. Change "#defines" below (or see below how to use COPYQ_INNO_SETUP environment variable).
 ; 3. Compile "setup.exe".
 
 ; Path for output installation file
-#define Output "."
+#define Output         "."
+#define MyAppName      "CopyQ"
+#define MyAppNameMin   "copy"
+#define MyAppCopyright "Lukas Holecek"
 
 [Setup]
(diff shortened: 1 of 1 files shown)


shared/copyq.iss
```diff
@@ -1,38 +1,49 @@
-﻿; 1. Open this file with Inno Setup with Unicode support and preprocessor.
+; 1. Open this file with Inno Setup with Unicode support and preprocessor.
 ; 2. Change "#defines" below (or see below how to use COPYQ_INNO_SETUP environment variable).
 ; 3. Compile "setup.exe".
 
 ; Path for output installation file
-#define Output "."
+#define Output         "."
+#define MyAppName      "CopyQ"
+#define MyAppNameMin   "copy"
```
hluk:
```suggestion
#define MyAppNameMin   "copyq"
```

hluk (commented):
Thanks for the fixes!

I believe the starting year is 2009.

shared/copyq.iss
bovirus: Tnahkls for the report about ypo error

bovirus: I updated the installer script with "MyAppCopyrightStartYear" and "MyAppCopyrighTEndYear variables".

Please check it. Thanks.

hluk:
Merged, thanks!

PR #2263, opened 2023-02-11
Update copyq.iss

VenusGirl:

PR #2265, opened 2023-02-13
Update copyq.iss

VenusGirl:


shared/copyq.iss
```diff
@@ -38,7 +38,6 @@ DefaultGroupName={#MyAppName}
 AllowNoIcons=yes
 LicenseFile={#Source}\LICENSE
 OutputDir={#Output}
-PrivilegesRequiredOverridesAllowed=dialog
```
hluk, line 41: Was this removed by mistake? (And also the empty lines below.)

hluk (changes requested):
Looks like some lines where removed by mistake.

Also see my previous comment: https://github.com/hluk/CopyQ/pull/2263#issuecomment-1426955551

BTW, it's better to keep existing PR open and just amend a commit (or add new commit) and `git push --force` to your branch again.

PR #2381, opened 2023-06-24
Update faq.rst for KeePassXC

axelson:
Make it clear that in addition to setting `format=x-kde-passwordManagerHint` you need to set clear `window`, otherwise the command will not match the KeePassXC application.


hluk (approved):
Thanks.

Issue #1208, opened 2019-08-31
copyq.sec missing on Windows version of Copyq


PEG71:
**Describe the bug**
Only copyq.pub is created when activating encryption with the Windows version of Copyq. No copyq.sec. Encryption is working despite the missing file.

**To Reproduce**
1. go to File, Preferences, Item, Encryption
2. Click on "generate new keys"
3. Enter passe phrase

**Expected behavior**
I'm expecting to find two key files in the %appdata%\copyq folder: copyq.pub and copyq.sec

**Version, OS and Environment**
Copyq version 3.9.2
Windows 10 1903
gpg4win 3.1.10

**Additional context**
I also get this error message when generating the key: gpg keydb_search failed: Permission denied.

hluk:
Looks like a bug in `gpg` utility that comes with that specific version of gpg4win.

Also, the encryption/decryption works because the `.pub` file contains both secret and public keys.

realchrisolin:
Seeing this same issue in v3.10, gpg4win 3.1.11, and the same Win10 build. copyq.pub is being created, but copyq.sec isn't.

```ERROR [2020-02-12 14:14:16.727] : ItemEncrypt ERROR: GnuPG stderr:
ERROR [2020-02-12 14:14:16.727] :    gpg: can't open 'C:\Users\chris.olin\AppData\Roaming\copyq\copyq.sec': No such file or directory
ERROR [2020-02-12 14:14:16.727] :    gpg: Total number processed: 0

**Comment by realchrisolin** [comment]
I did some troubleshooting and found that gpg wasn't generating the secret key in both the Windows **and** Linux builds. Turns out that changes in v2.1+ in how secret keys are stored ultimately broke the `%secring` control statement. See https://superuser.com/a/1037402, https://www.gnupg.org/faq/whats-new-in-2.1.html#nosecring, and https://www.gnupg.org/documentation/manuals/gnupg/Ephemeral-home-directories.html. 

Suffice to say, `startGenerateKeysProcess()` and probably other parts of the encryption plugin will need to be rewritten from scratch. However, it looks like I was able to get encryption working by turning off the encryption plugin, quitting copyq, deleting any .pub or pub.lock files in the copyq config directory, interactively creating a RSA encrypt key, manually exporting public and secret keys, then finally copying them into the copyq config directory, launching copyq, and turning encryption back on. I was prompted for the secret key passphrase and am no longer seeing errors in the log.

```
[...]

PR #2464, opened 2023-09-14
update ItemEncrypted plugin to support GPG 2.1

### LINKED ISSUE #2463: Rewrite ItemEncrypted plugin to support modern versions of GPG
Related to #1208 and #2363, the ItemEncrypted plugin is incompatible with modern versions of GPG and the only workaround is to use an outdated version that is several years old.

This ticket exists to explicitly call out the ItemEncrypted plugin as the culprit and document/schedule/track progress on this front. The crux of this issue is that the modern GPG executable doesn't support all of the (legacy) arguments, exact details [documented in my comment three years ago](https://github.com/hluk/CopyQ/issues/1208#issuecomment-585442955).

I would rewrite this, but my experience with C++ and Qt is limited. I understand enough to read the relevant code, but I'm not familiar enough with either to know where to start without a lot of hours of research first.

realchrisolin:
This PR is a work-in-progress designed to resolve #2463. I don't know if I'll ever be able to finish it, but it should give someone more familiar with C++ and Qt that's trying to fix this issue a place to start and build on.

### DIFF
--- a/plugins/itemencrypted/itemencrypted.cpp
+++ b/plugins/itemencrypted/itemencrypted.cpp
@@ -116,7 +116,7 @@ bool checkGpgExecutable(const QString &executable)
 
 bool checkGpgVersionOutput(const QString &executable)
 {
-    const auto versionOutptput(executable);
+    const auto versionOutput = getGpgVersionOutput(executable);
     QStringList lines = versionOutput.split('\n');
     QString firstLine = lines.first();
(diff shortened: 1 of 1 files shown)


hluk:
Actually, I think I understand the situation from your old comment (https://github.com/hluk/CopyQ/issues/1208#issuecomment-585442955). Let me try implementing the fix.

hluk:
Should be fixed - see https://github.com/hluk/CopyQ/issues/1208#issuecomment-1728217576.

Thanks for the help! :pray:

PR #2508, opened 2023-10-12
itemfakevim: fix build with qt 6.6.0

NickCao:
Reference: https://github.com/qt-creator/qt-creator/commit/e56e3b6f374e00179eb0537198437864dddc47f2


hluk:
Thanks for the fix!

PR #2588, opened 2024-01-09
Add codespell config + workflow and make it fix some typos

yarikoptic:
My humble contribution -- I have been a "heavy" (albeit really basic) user of CopyQ for years. It is a great productivity tool -- thank you very much for it!!!

### DIFF
--- a/plugins/itemfakevim/fakevim/fakevimhandler.cpp
+++ b/plugins/itemfakevim/fakevim/fakevimhandler.cpp
@@ -864,7 +864,7 @@ static QString getProcessOutput(const QString &command, const QString &input)
     proc.write(toLocalEncoding(input));
     proc.closeWriteChannel();
 
-    // FIXME: Process should be interruptable by user.
+    // FIXME: Process should be interruptible by user.
     //        Solution is to create a QObject for each process and emit finished state.
     proc.waitForFinished();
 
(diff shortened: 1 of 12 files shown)


plugins/itemfakevim/fakevim/fakevimhandler.cpp
```diff
@@ -6504,7 +6504,7 @@ bool FakeVimHandler::Private::handleExMultiRepeatCommand(const ExCommand &cmd)
 {
     // :[range]g[lobal]/{pattern}/[cmd]
     // :[range]g[lobal]!/{pattern}/[cmd]
-    // :[range]v[globa]!/{pattern}/[cmd]
+    // :[range]v[global]!/{pattern}/[cmd]
```
yarikoptic, line 6507: is this a false positive?

hluk:
Oh, cool, this looks very handy. Thanks!

plugins/itemfakevim/fakevim/fakevimhandler.cpp
hluk: I don't think so. This file is actually taken from Qt Creator, but there are some incompatible changes so it may take a while before I sync the changes again.

hluk (approved):

PR #2609, opened 2024-02-05
macOS: Add M1/arm64 support

### LINKED ISSUE #1884: Provide native(universal) builds for M1(arm64) MacBooks
**Is your feature request related to a problem? Please describe.**
I installed the latest build of CopyQ from Releases but realised that it's still only compiled for x86 architecture. I can still use it almost perfectly with the rosetta translation of the M1 chip but nothing like having a native build.

**Describe the solution you'd like**
I'm not sure if we already can successfully compile for the M1(arm64), but it would be great if you guys can provide an universal macOS package with native binaries for M1. I'm more than happy if there is anything I can help with in terms of testing and such :)

**Describe alternatives you've considered**
_NULL_

**Additional context**
Nothing, but thanks for reading!

hluk:
To fix an issue with deactivates app when returning focus to the main window after closing a dialog, the main window type is no longer set explicitly (as `Qt::Sheet` with `Qt::WindowModal` modality). This may again disallow opening the window on top of fullscreen apps. A workaround to allow easier testing in the future could be to make this configurable.

Closes #1884


hluk:
There is a problem with dialog windows not being focused in tests. I tried disabling animations but without any luck.

PR #2800, opened 2024-08-04
Update OS X build-from-source doc to using Qt6 by default

Gudauu:
Reason for this PR:
* The current OS X build dependency specified in the `Homebrew` formulae defaults to `Qt6` instead of `Qt5`. Therefore, `Qt6` should be used as the default in `CMake` commands and docs.

What changed:

* In `CMakeLists.txt`, change `WITH_QT6` to `ON`.
* Update the OS X build-from-source docs accordingly.

Tests:

* Tested on macOS 14.5; the build succeeded without problem.

Behaviour before this change:

* Followed docs precisely, build failed at the `CMake` stage with error: “Could not find a package configuration file provided by Qt5Widgets.”

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,7 @@ endif()
 
 set(CMAKE_AUTOMOC ON)
 
-OPTION(WITH_QT6 "Enable Qt 6" OFF)
+OPTION(WITH_QT6 "Enable Qt 6" ON)
 if (WITH_QT6)
     set(QT_DEFAULT_MAJOR_VERSION 6)
     find_package(Qt6Widgets 6.1.0 REQUIRED)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ endif()
 
 set(CMAKE_AUTOMOC ON)
 
-OPTION(WITH_QT6 "Enable Qt 6" OFF)
+OPTION(WITH_QT6 "Enable Qt 6" ON)
```
hluk, line 46: This changes using Qt 6 by default everywhere and would break a couple of builds for Linux distros.

Would it be possible to make this automatic? Specifically:
- By default if the option is not overridden, pick Qt 6 preferably and if not found fallback to Qt 5.
- If set to ON by user, require Qt 6.
- If set to OFF by user, require Qt 5.

hluk:
Thanks for the patch! Most distros should now use Qt 6 (AFAIK there has not been a Qt 5 release for a while) - except perhaps long-term-support Ubuntu releases.

PR #2829, opened 2024-08-30
Add Emacs navigation key-bindings support

janmejay:
Rebase and address review-comments on a very old abandoned change (PR#1039) 
* add emacs-style navigation keymap support (similar to existing Vi keymap support)
* convert the checkboxes to a set of radio-buttons (one of the open review comments)

### DIFF
--- a/src/common/common.cpp
+++ b/src/common/common.cpp
@@ -726,6 +726,71 @@ bool handleViKey(QKeyEvent *event, QObject *eventReceiver)
     return true;
 }
 
+bool handleEmacsKey(QKeyEvent *event, QObject *eventReceiver)
+{
+    int key = event->key();
+    Qt::KeyboardModifiers mods = event->modifiers();
+
+    switch ( key ) {
+    case Qt::Key_V:
+        if (mods & Qt::ControlModifier) {
+            key = Qt::Key_PageDown;
+            mods = mods & ~Qt::ControlModifier;
+        } else if (mods & Qt::AltModifier) {
+            key = Qt::Key_PageUp;
+            mods = mods & ~Qt::AltModifier;
+        } else {
+        return false;
+    }
+    break;
+    case Qt::Key_N:
+        if (mods & Qt::ControlModifier) {
+            key = Qt::Key_Down;
+            mods = mods & ~Qt::ControlModifier;
+            break;
+        }
+        return false;
+    case Qt::Key_P:
+        if (mods & Qt::ControlModifier) {
+            key = Qt::Key_Up;
+            mods = mods & ~Qt::ControlModifier;
+            break;
+        }
+        return false;
+    case Qt::Key_Less:
+        if ((mods & Qt::AltModifier)) {
+            key = Qt::Key_Home;
+            mods = mods & ~(Qt::ShiftModifier | Qt::AltModifier);
+			break;
+        } else {
+            return false;
+        }
+    case Qt::Key_Greater:
+        if ((mods & Qt::AltModifier)) {
+            key = Qt::Key_End;
+            mods = mods & ~(Qt::ShiftModifier | Qt::AltModifier);
+			break;
+        } else {
+            return false;
+        }
+    case Qt::Key_G:
+        if (mods & Qt::ControlModifier) {
+            key = Qt::Key_Escape;
+            mods = mods & ~Qt::ControlModifier;
+            break;
+        }
+        return false;
+    default:
+        return false;
+    }
+
+    QKeyEvent event2(QEvent::KeyPress, key, mods, event->text());
+    QCoreApplication::sendEvent(eventReceiver, &event2);
+    event->accept();
+
+    return true;
+}
+
 bool canDropToTab(const QDropEvent &event)
 {
     const auto &data = *event.mimeData();
--- a/src/common/common.cpp
+++ b/src/common/common.cpp
@@ -736,12 +736,14 @@ bool handleEmacsKey(QKeyEvent *event, QObject *eventReceiver)
         if (mods & Qt::ControlModifier) {
             key = Qt::Key_PageDown;
             mods = mods & ~Qt::ControlModifier;
-        } else if (mods & Qt::AltModifier) {
+            break;
+        }
+        if (mods & Qt::AltModifier) {
             key = Qt::Key_PageUp;
             mods = mods & ~Qt::AltModifier;
-        } else {
+            break;
+        }
         return false;
-    }
     break;
     case Qt::Key_N:
         if (mods & Qt::ControlModifier) {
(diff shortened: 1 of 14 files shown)


janmejay:
@hluk did you have a chance to look at this?

hluk:
> @hluk did you have a chance to look at this?

Not yet. Thanks for the patch! Checking now.

src/common/common.cpp
```diff
@@ -726,6 +726,71 @@ bool handleViKey(QKeyEvent *event, QObject *eventReceiver)
...
+        }
+        return false;
+    case Qt::Key_Less:
+        if ((mods & Qt::AltModifier)) {
+            key = Qt::Key_Home;
+            mods = mods & ~(Qt::ShiftModifier | Qt::AltModifier);
+			break;
+        } else {
+            return false;
+        }
```
hluk:
```suggestion
            break;
        }
        return false;
```

src/common/common.cpp
```diff
@@ -726,6 +726,71 @@ bool handleViKey(QKeyEvent *event, QObject *eventReceiver)
...
+    case Qt::Key_Greater:
+        if ((mods & Qt::AltModifier)) {
+            key = Qt::Key_End;
+            mods = mods & ~(Qt::ShiftModifier | Qt::AltModifier);
+			break;
+        } else {
+            return false;
+        }
```
hluk:
```suggestion
            break;
        }
        return false;
```

src/common/common.cpp
```diff
@@ -726,6 +726,71 @@ bool handleViKey(QKeyEvent *event, QObject *eventReceiver)
...
+    case Qt::Key_V:
+        if (mods & Qt::ControlModifier) {
+            key = Qt::Key_PageDown;
+            mods = mods & ~Qt::ControlModifier;
+        } else if (mods & Qt::AltModifier) {
+            key = Qt::Key_PageUp;
+            mods = mods & ~Qt::AltModifier;
+        } else {
+        return false;
+    }
```
hluk: To keep the code structure consistent:

```suggestion
            mods = mods & ~Qt::ControlModifier;
            break;
        }
        if (mods & Qt::AltModifier) {
            key = Qt::Key_PageUp;
            mods = mods & ~Qt::AltModifier;
            break;
        }
        return false;
```

src/gui/mainwindow.cpp
```diff
@@ -2557,6 +2557,19 @@ void MainWindow::keyPressEvent(QKeyEvent *event)
...
+        if ((modifiers == Qt::ControlModifier && key == Qt::Key_G)
+            || (key == Qt::Key_Escape)) {
+            onEscape();
+            return;
+        }
+
+        if (browseMode()) {
+            if (c && handleEmacsKey(event, c))
+                return;
+        }
```
hluk:
```suggestion
        if (browseMode() && c && handleEmacsKey(event, c)) {
                return;
        }
```

src/gui/traymenu.h
```diff
@@ -83,6 +86,7 @@ class TrayMenu final : public QMenu
 
     bool m_omitPaste;
     bool m_viMode;
+    bool m_emacsMode;
```
hluk: Can you change this to use an enum? Something like:

```cpp
enum class NavigationStyle {
    Default,
    Vi,
    Emacs
};
```

It would also mean that there is single method to set the navigation `void setNavigationStyle(NavigationStyle style)`.

It would be best if the configuration value in `appconfig.h` has this type, but that might require some more refactoring. I could look into this later.

hluk (changes requested):

src/gui/traymenu.h
janmejay: @hluk I looked into this today. One way I can see of making this change (let me know if this is not really what you had in mind) is to make `NavigationStyle` a `QVariant` and then bind `QGroupBox` instead of individual `QRadioButtons`.

I'd have to overload the `Option` constructor to allow me to pass getter and setter lambdas because `checkedId` is not available as a property. This overload would be used for `Option<NavigationStyle>` instance.

The main problem is this is backwards incompatible with persisted settings. For anyone that has `vi` mode enabled at the moment would reset to `Default` when they upgrade.

We can try to migrate settings, but unless we introduce a new concept (say `SettingsMigrator`) the implementation won't be clean.

Let me know if this isn't what you had in mind.

hluk: I would prefer to keep the code simpler. I tried writing some conversion function for new enum option, but it did not work well and was unnecessarily complex (custom QMetaType conversion function).

Here is a simple solution: https://github.com/hluk/CopyQ/commit/28100299af198a9f412e86e9328481adb2a5c944

The UI uses QComboBox with fixed values instead of a group of radio buttons - enum values and the items in the combo box must match.

It reuses the `Config::vi` option because I thought that it would convert the old `false`/`true` values so it would stay backwards-compatible. Unfortunately, that does not work. It would be good to figure this out, but I would avoid any complex migration code.

Another downside is that from CLI (`copyq config vi ...`) you would need to use values 0,1,2 instead of more reasonable strings (e.g. "vi"/"emacs").

hluk: I've solved the backwards-compatibility issue in: https://github.com/hluk/CopyQ/commit/aba654f6698e5ba16579c02bd2538415301f6def

But I'm not sure if I like the complexity of the solution 🙂 ([diff from the previous version](https://github.com/hluk/CopyQ/compare/28100299af198a9f412e86e9328481adb2a5c944..aba654f6698e5ba16579c02bd2538415301f6def)).

janmejay: @hluk you actually did all the work, I applied https://github.com/hluk/CopyQ/commit/aba654f6698e5ba16579c02bd2538415301f6def and doesn't look like there is anything left to do. 

I had a `predicate -> transform` arrangement in mind for migrating config over, that is, if `predicate` is true apply `transform`. So predicate would look for `vi=true`, remove it and replace with `navigation_style=vi` etc, but I have never worked with Qt before this, so not sure whether this would work well with `QSettings`.

If you prefer avoiding migration altogether, I can pull https://github.com/hluk/CopyQ/commit/28100299af198a9f412e86e9328481adb2a5c944 into this PR (and perhaps change it from `Config::vi` to `Config::navigation_style`).

What are the next steps you have in mind?

hluk: > If you prefer avoiding migration altogether, I can pull [2810029](https://github.com/hluk/CopyQ/commit/28100299af198a9f412e86e9328481adb2a5c944) into this PR (and perhaps change it from `Config::vi` to `Config::navigation_style`).

That sounds good, also with changing the name of the struct, and perhaps even changing the name of the option from "vi" to "navigation_style" to avoid confusion.

janmejay:
@hluk Sorry for the delay, couldn't get to this until today. Refreshed the PR.

hluk:
@janmejay Thanks for the feature! :+1:

PR #2933, opened 2024-12-16
Update copyq.desktop

alibij:
fix run app in ubuntu 24LTS


hluk:
I would prefer not to force the app to run in X11 compatibility mode on Wayland. There are other problems with this, like wrong scaling or blurred UI, and the clipboard may not work anyway if you copy from Wayland apps.

I run the application in natively on Wayland (mainly KDE and Sway) without any big problems.

PR #3070, opened 2025-05-30
Docs: Remove invalid www from flatpak links

allella:
The docs have links using the _www_ subdomain, but Flatpak does not use or redirect _www_.

This causes browser certificate warnings when trying to click from the CopyQ docs to Flatpak's site.

It seems [unlikely Flatpak will add support for a www subodmain](https://discourse.flathub.org/t/www-flathub-org-isnt-301ing-to-flathub-org/7941).

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Install `copyq` package.
 
 ### Other Linux Distributions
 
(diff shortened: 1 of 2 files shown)


hluk:
Thanks for the fix!

PR #3115, opened 2025-07-27
Added doc for copy() problem & workaround

db2222:
As several people have run into this issue added this doc.

Also found a workaround for Windows which probably could be helpful to others.

### DIFF
--- a/docs/known-issues.rst
+++ b/docs/known-issues.rst
@@ -124,8 +124,9 @@ An alternative under Windows is to use a Powershell script as a command:
 
     powershell:
       Add-Type -AssemblyName System.Windows.Forms;
+      Start-Sleep -Milliseconds 300;
       [System.Windows.Forms.SendKeys]::SendWait("^c");
-      Start-Sleep -Milliseconds 150;
+      Start-Sleep -Milliseconds 300;
 
 The delay is added to make sure the text is copied to the clipboard.
 
--- a/docs/known-issues.rst
+++ b/docs/known-issues.rst
@@ -119,15 +119,22 @@ This can fail depending on the active application.
 If CopyQ won't detect a clipboard change, it throws an exception.
 The execution then fails with the message ``Failed to copy to clipboard!``.
 
-An alternative under Windows is to use a Powershell script as a command:
-  .. code-block:: powershell
+An alternative under Windows is to use a Powershell script to override the ``copy`` operation
+(see :ref:`faq-share-commands`):
 
-    powershell:
-      Add-Type -AssemblyName System.Windows.Forms;
-      Start-Sleep -Milliseconds 300;
-      [System.Windows.Forms.SendKeys]::SendWait("^c");
-      Start-Sleep -Milliseconds 300;
+.. code-block:: powershell
 
-The delay is added to make sure the text is copied to the clipboard.
+    [Command]
+    Command="
+        copy = function() {
+            execute('powershell', '-Command', `
+                Add-Type -AssemblyName System.Windows.Forms;
+                Start-Sleep -Milliseconds 300;
+                [System.Windows.Forms.SendKeys]::SendWait(\"^c\");
+                Start-Sleep -Milliseconds 300;
+            `);
+        }"
+    IsScript=true
+    Name=Override copy()
 
-Beware that necessary **permissions** for Powershell independent of CopyQ need to have been setup.
+The delays are added to make sure no focus issues occur and the text is copied to the clipboard.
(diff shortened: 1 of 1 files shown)


docs/known-issues.rst
```diff
@@ -110,3 +110,23 @@ For **Flatpak** application, see `this workaround
...
+
+Scripting command "copy()" fails
+--------------------------------
+
+The command ``copy()`` sends the Ctrl+C shortcut to the current window.
+This can fail depending on the active application.
+If CopyQ won't detect a clipboard change, it throws an exception.
+The execution then fails with the message ``Failed to copy to clipboard!``.
+
+An alternative under Windows is to use a Powershell script as a command:
```
hluk: Thanks for the Powershell script!

How do you use this?

It could be more useful to override `copy()` (using a "Script" command type) so it can be used by all command out-of-the-box:

```ini
[Command]
Command="
    copy = function() {
        execute('powershell', '-Command', `
            Add-Type -AssemblyName System.Windows.Forms;
            [System.Windows.Forms.SendKeys]::SendWait(\"^c\");
            Start-Sleep -Milliseconds 150;
        `);
    }"
Icon=
IsScript=true
Name=Override copy()
```

docs/known-issues.rst
```diff
@@ -110,3 +110,23 @@ For **Flatpak** application, see `this workaround
...
+  .. code-block:: powershell
+
+    powershell:
+      Add-Type -AssemblyName System.Windows.Forms;
+      [System.Windows.Forms.SendKeys]::SendWait("^c");
+      Start-Sleep -Milliseconds 150;
+
+The delay is added to make sure the text is copied to the clipboard.
+
+Beware that necessary **permissions** for Powershell independent of CopyQ need to have been setup.
```
hluk: Can you provide more details? Some external link or simple description how to set up the permissions.

docs/known-issues.rst
db2222: I use it to copy a Jira ticket number from anywhere (i.e. text editor) and open it directly in the browser via one shortcut (Win+J).

Interesting. I wasn't aware that this was possible with CopyQ and therefore wrote a PowerShell script that does it all in one go. Your solution is probably better suited for the documentation as it solves any usage of the `copy()` command.

docs/known-issues.rst
db2222: It's a bit tricky to give a general answer.
It can depend on the execution restrictions from the IT of a company for instance.
My intention was to hint in the direction so that nobody comes to CopyQ with "not working" complaints but in reality PowerShell isn't allowed/setup 😉
For instance see these links:
- https://superuser.com/questions/106360/how-to-enable-execution-of-powershell-scripts
- https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.security/set-executionpolicy?view=powershell-7.5

db2222:
@hluk Is anything else needed from me? 🙂
Should I apply your script variant?

docs/known-issues.rst
hluk: Perhaps better to remove this, unless a specific problem is really expected to occur and you can provide more details.

docs/known-issues.rst
```diff
@@ -110,3 +110,24 @@ For **Flatpak** application, see `this workaround
...
+The execution then fails with the message ``Failed to copy to clipboard!``.
+
+An alternative under Windows is to use a Powershell script as a command:
+  .. code-block:: powershell
+
+    powershell:
+      Add-Type -AssemblyName System.Windows.Forms;
+      Start-Sleep -Milliseconds 300;
+      [System.Windows.Forms.SendKeys]::SendWait("^c");
+      Start-Sleep -Milliseconds 300;
```
hluk:
```suggestion
An alternative under Windows is to use a Powershell script to override copy operation
(see :ref:`faq-share-commands`):

  .. code-block:: powershell

    [Command]
    Command="
        copy = function() {
            execute('powershell', '-Command', `
                Add-Type -AssemblyName System.Windows.Forms;
                [System.Windows.Forms.SendKeys]::SendWait(\"^c\");
                Start-Sleep -Milliseconds 150;
            `);
        }"
    Icon=
    IsScript=true
    Name=Override copy()
```

docs/known-issues.rst
db2222: I have removed it.

docs/known-issues.rst
db2222: I have changed it but slightly modified.
The formatting needs to be different or Sphinx complains.
I haven't found a way to add the ```Icon``` line without having a warning so I left it out.
In general I have never worked with Sphinx before, so I would have to investigate too much time.
Also I took my code with the 2 delays as during testing a few weeks ago this variant is actually reliable unlike the previous one.

hluk:
Thanks! :+1:

PR #3164, opened 2025-09-11
Add wastebin feature for creating burner secrets

thereisnotime:
This update introduces a new feature that allows users to create burner secrets in a wastebin. It includes the addition of configuration options for enabling the feature and specifying the wastebin endpoint. The context menu is updated to include an action for creating a wastebin secret, and the necessary logic for handling uploads and responses is implemented in the main window.

Key changes:
- Added `wastebin_enabled` and `wastebin_endpoint` configuration options.
- Implemented `createWastebinSecret`, `uploadToWastebin`, and `handleWastebinResponse` methods in `MainWindow`.
- Updated UI to include controls for the wastebin feature in the general configuration tab.
- Added context menu action for creating a burner secret in the wastebin.

### DIFF
--- a/src/common/appconfig.h
+++ b/src/common/appconfig.h
@@ -543,6 +543,22 @@ struct frameless_window : Config<bool> {
     }
 };
 
+struct wastebin_enabled : Config<bool> {
+    static QString name() { return QStringLiteral("wastebin_enabled"); }
+    static Value defaultValue() { return false; }
+    static const char *description() {
+        return "Enable 'Create burner secret in wastebin' feature in context menu";
+    }
+};
+
+struct wastebin_endpoint : Config<QString> {
+    static QString name() { return QStringLiteral("wastebin_endpoint"); }
+    static Value defaultValue() { return QStringLiteral("https://bin.rso.dev"); }
+    static const char *description() {
+        return "Wastebin endpoint URL for uploading burner secrets";
+    }
+};
+
 } // namespace Config
 
 class AppConfig final
--- a/src/gui/configurationmanager.cpp
+++ b/src/gui/configurationmanager.cpp
@@ -344,6 +344,9 @@ void ConfigurationManager::initOptions()
     bind<Config::close_on_unfocus_delay_ms>();
 
     bind<Config::frameless_window>();
+
+    bind<Config::wastebin_enabled>(m_tabGeneral->checkBoxWastebinEnabled);
+    bind<Config::wastebin_endpoint>(m_tabGeneral->lineEditWastebinEndpoint);
 }
 
 template <typename Config, typename Widget>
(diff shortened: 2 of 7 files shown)


hluk:
Not entirely sure what are "burner secrets" or "wastebin", but quick scan of the code shows that it does some network requests. This is not something that should be part of CopyQ core - the app doesn't do any remote requests by default.

It would be better to implement it using commands/scripting.

thereisnotime:
@hluk 

I understand your concern about network requests in core. The "wastebin" feature is conceptually similar to how screenshot tools (like Lightshot or Flameshot) allow direct upload to image hosts. Instead of forcing the user to copy content, open a browser, paste, and then share a link, CopyQ could streamline that workflow for text secrets.

Why it’s useful:

Speed & UX: One click “burner secret” -> instant shareable link, just like image uploaders.

Security: Unlike storing in persistent pastebins, wastebin endpoints are temporary and auto-expiring by design (hence “burner”).

Consistency: Many users already rely on CopyQ for quick sharing of snippets; this adds a natural extension.

Why core (not just scripts):

Scriptability is powerful, but discoverability and smooth UX suffer. A built-in menu item makes it approachable for non-technical users.

Similar precedent: clipboard managers with screenshot integrations already exist. This does not expand CopyQ’s scope beyond “clipboard + quick share”.

To address concerns, it could be:

Optional and opt-in (disabled by default, requires user-configured endpoint).

Minimal core impact (isolated feature, no background network requests).

That way, you keep CopyQ’s local-first philosophy but provide a built-in integration point for those who want frictionless secret sharing.

hluk:
I still don't see much reason not to create a command instead.

The downsides of having this in the application:
- it is more code to maintain (it also needs tests)
- it is hardcoded into the mainwindow module (which itself already needs to be refactored)
- I've never heard about "bin.rso.dev" server and could not find any information about it (Kagi search failed)

I prefer to keep the core application as simple as possible while still being easily extendable.

> Speed & UX: One click “burner secret” -> instant shareable link, just like image uploaders.

I think the command would not be much slower, and it also allows you to create a simple menu item.

> Security: Unlike storing in persistent pastebins, wastebin endpoints are temporary and auto-expiring by design (hence “burner”).

To me, this is something that would make the app less secure, since it allows sending data to some server with a single click.

> Consistency: Many users already rely on CopyQ for quick sharing of snippets; this adds a natural extension.

Contrary to what you say: this seems to me like a very specific use case not many clipboard manager users may care about. I code daily, but almost never need to share code snippets through some 3rd party websites.

> Scriptability is powerful, but discoverability and smooth UX suffer. A built-in menu item makes it approachable for non-technical users.

A good point, but it would be better to improve the way users can extend the application and discover custom commands.

thereisnotime:
@hluk will scripting allow me to add a custom menu command in the history list?

hluk:
> will scripting allow me to add a custom menu command in the history list?

Yes. See: https://copyq.readthedocs.io/en/latest/writing-commands-and-adding-functionality.html

See also: https://copyq.readthedocs.io/en/latest/scripting-api.html

PR #3190, opened 2025-09-20
Linux: Fix selection synchronization; avoid error on SIGINT and fix piping data in sub-processes

hluk:


robertstrom:
question - is this fix related to an issue that I posted where I would chose a particular item to paste and CopyQ would paste the item before the item selected?

hluk:
> question - is this fix related to an issue that I posted where I would chose a particular item to paste and CopyQ would paste the item before the item selected?

@robertstrom I don't think so. It mainly fixes some issues in unreleased code.

PR #3204, opened 2025-09-26
Tests: Add tests drag'n'drop items to tab

hluk:


hluk:
Dropping. It does not look like QTest can handle drag'n'drop properly between multiple widgets.

PR #3268, opened 2025-10-22
when Qt6 is used make sure cmake variables are there

BlackIkeEagle:
When using `WITH_QT6` use
`find_package(QT REQUIRED COMPONENTS Core NAMES Qt6)` to make sure the QT related cmake variables are there. There is a use of the variable `QT_VERSION` further down the line while checking for Qt::GuiPrivate in the `src/platform/x11/x11platform.cmake` file.

This change makes that check work properly so we can build with Qt6 >= 6.10.


hluk:
Thanks for the fix!

texstar:
Thank you. You're the best!

PR #3303, opened 2025-11-07
Add nord themes

hedgieinsocks:
Hello, if you are open to accepting these themes based on https://www.nordtheme.com/ palette, here they are. I'm not a frontend/css guy at all, so css code probably might be improved since it was created manually after inspecting the source code and existing themes (no AI).   

* dark

image

* light

image

P.S. I wanted to make it roundish to fit GNOME aesthetics but alas I could not solve white corners bug https://github.com/hluk/CopyQ/discussions/3292 so it is squarish for now.

### DIFF
--- a/shared/themes/nord.ini
+++ b/shared/themes/nord.ini
@@ -1,0 +1,127 @@
+[General]
+css_template_items=items
+css_template_main_window=main_window
+css_template_menu=menu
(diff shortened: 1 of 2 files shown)


hluk:
Very cool! Thanks!

PR #3317, opened 2025-11-13
don't start the monitoring after stopping it

### LINKED ISSUE #3307: kde hangs at logout for a few seconds when copyq is running
**Describe the bug**
kde hangs at logout for about 5 seconds when copyq is running

**To Reproduce**
- start copyq
- exit kde
- screen stays black for about 5 seconds

**Expected behavior**
kde exits instantly

CopyQ Clipboard Manager 13.0.0
Qt: 6.9.3
Compiler: GCC
Arch: x86_64-little_endian-lp64
OS: Linux

puchu:
Fixes #3307

### DIFF
--- a/src/app/clipboardserver.cpp
+++ b/src/app/clipboardserver.cpp
@@ -586,7 +586,6 @@ void ClipboardServer::onMonitorFinished()
     if (!m_monitor)
         setClipboardMonitorRunning(false);
     stopMonitoring();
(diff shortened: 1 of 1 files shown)


hluk:
This is needed to restart the clipboard monitor in case it exits itself or crashes.

I think #3318 could fix the problem.

Thanks for finding the issue!

PR #3339, opened 2025-12-01
Clean up for Qt 6.10

BSDKaffee:
Clean up after e061ca9 and 1039035 because of e7dd572. So, e7dd572 should have included a Qt version check. e061ca9 adds this check and fixes build with Qt <= 6.9, but there is no QT_VERSION variable defined in the Qt CMake configs. This again breaks Qt >= 6.10. So, 1039035 does some creative things to make the build work again with Qt 6.10, but this is misguided and non-standard. All that needs be done is to compare Qt6_VERSION (case sensitive) in src/platform/x11/x11platform.cmake.


hluk:
Thank you!

Issue #3382, opened 2026-01-12
Missing dependency on Ubuntu


QuentiumYT:
**Describe the bug**
Cannot install libkf6guiaddons & libkf6guiaddons-dev on Ubuntu LTS (noble)
None exist in Ubuntu repos, only Kubuntu have them. How can we build it on Ubuntu?

**Version, OS and Environment**

- Application Version (for example 10.0.0)
- OS Mint 22.2
- Desktop environment, window manager Cinnamon
- Installation method from source

Chealer:
~~This is an ITS. Unless this ticket intends to report an issue, please discard it and use [the mailing list](https://groups.google.com/g/copyq) instead.~~

In any case, you should indicate the version you use. And these packages do exist in Ubuntu repositories, just not in your old version. Ubuntu 24.04 does not include KDE Frameworks 6.

QuentiumYT:
I haven't seen any related ticket, my bad if one exists.
I just read about this statement https://groups.google.com/g/copyq/c/KElH_U2hMO0, so maybe that's why Ubuntu 24.04 will no longer be supported, meaning it's already "old" regarding your answer. libkf6guiaddons exists since Ubuntu plucky (25.04) so does that mean "build from source" will not work for older releases?

Chealer:
I am sorry, I may have misunderstood your report, but in any case, you should indicate the CopyQ version you use. And why are you trying to install from source?

<details>
  <summary>License</summary>
This comment and any other from myself on this ticket is offered under the terms of CC0 1.0.
</details>

QuentiumYT:
No problem, I am not using any CopyQ version because I try to build it from source to use Qt6. I didn't want to install all Qt5 requirements only for this app and look into newer features :)

Chealer:
OK, but which CopyQ version are you trying to build? Are you following a compilation procedure?

QuentiumYT:
Sorry, I build from master.
I looked into the code and added `-DWITH_NATIVE_NOTIFICATIONS=false` which doesn't require libkf6guiaddons

Chealer:
Ah, I see, thanks. It would probably be best for you to use the code from the latest release instead, but I can see in [the documentation](https://copyq.readthedocs.io/en/latest/build-source-code.html#ubuntu) that there is indeed an issue with Ubuntu 24.04, which must indeed affect your Linux Mint version, so you found a good workaround.

Please give this ticket a more specific title. I suggest:
[...]

PR #3385, opened 2026-01-14
Update build from source docs on previous Ubuntu versions for KF6

QuentiumYT:
Hello, I stumbled upon a compilation problem, more context in #3382

This PR will set both KF6 Notifications & StatusNotifierItem optional because they are not available on Ubuntu LTS. Only Ubuntu >=25.04 (https://packages.ubuntu.com/plucky/libkf6guiaddons-dev) have it.

Compilation works fine without the flag `WITH_NATIVE_NOTIFICATIONS`, it just disable native notification support, that could only be available on recent versions.
Also `src/platform/x11/x11platform.cmake` already handle that case where built-in clipboard support could be skipped if no kguiaddons are found.

Updated build from source docs and kguiaddons typo accordingly

### DIFF
--- a/docs/build-source-code.rst
+++ b/docs/build-source-code.rst
@@ -37,8 +37,6 @@ On **Ubuntu** you can install all build dependencies with:
       cmake \
       extra-cmake-modules \
       git \
-      libkf6guiaddons \
-      libkf6guiaddons-dev \
       libqca-qt6-2 \
       libqca-qt6-dev \
       libqca-qt6-plugins \
@@ -60,7 +58,15 @@ On **Ubuntu** you can install all build dependencies with:
       qt6-wayland-dev-tools \
       qtkeychain-qt6-dev
 
-Fedora / RHEL / Centos
+On Ubuntu 25.04 and newer, you can also install `KF6` packages:
+
+::
+
+    sudo apt install \
+      libkf6guiaddons \
+      libkf6guiaddons-dev \
+
+Fedora / RHEL / CentOS
 ^^^^^^^^^^^^^^^^^^^^^^
 On **Fedora** and derivatives you can install all build dependencies with:
 
@@ -97,6 +103,7 @@ Build the source code with CMake and make or using an IDE of your choice (see ne
 
     cd CopyQ
     cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local .
+    # Add -DWITH_NATIVE_NOTIFICATIONS=OFF for Ubuntu systems without `KF6` packages
     make
     make install
 
--- a/src/platform/x11/x11platform.cmake
+++ b/src/platform/x11/x11platform.cmake
@@ -59,7 +59,7 @@ if (WITH_QT6 AND ECM_VERSION VERSION_GREATER "6.2.0")
 else()
     message(WARNING
         "Using built-in clipboard support."
-        " Requires Qt 6 and 'kf6-guiaddons', 'libkf6guiaddons' or similar"
+        " Requires Qt 6 and 'kf6-kguiaddons', 'libkf6guiaddons' or similar"
         " for better Wayland clipboard integration.")
     set(CMAKE_MODULE_PATH ${ECM_MODULE_PATH})
     include_directories(${CMAKE_CURRENT_BINARY_DIR}/platform/x11/systemclipboard)


src/notifications.cmake
```diff
@@ -8,14 +8,23 @@ if (WITH_NATIVE_NOTIFICATIONS)
 
     if (WITH_QT6)
         set(KF6_MIN_VERSION "5.248.0")
-        find_package(KF6 ${KF6_MIN_VERSION} REQUIRED COMPONENTS Notifications StatusNotifierItem)
-        list(APPEND copyq_LIBRARIES KF6::Notifications KF6::StatusNotifierItem)
+        find_package(KF6 ${KF6_MIN_VERSION} COMPONENTS Notifications StatusNotifierItem)
+        if (KF6_FOUND)
+            list(APPEND copyq_LIBRARIES KF6::Notifications KF6::StatusNotifierItem)
+        else()
+            set(WITH_NATIVE_NOTIFICATIONS OFF CACHE BOOL "Build with native notification support" FORCE)
```
hluk: If a user sets the option explicitly, the build system should not revert it, that would be unexpected. I would also prefer if my automated builds just fail if some libraries are missing, instead of seeing unexpected behavior later.

QuentiumYT: I understand, the option is ON by default but cannot work on most Debian & Ubuntu users because the lib doesn't exist. What should we do? Also what is the behavior of this not required library in x11platform.cmake I'm not quite into it.

hluk: It needs to be disabled explicitly if the libraries cannot be installed: `cmake -DWITH_NATIVE_NOTIFICATIONS=OFF ....`

I don't have a better solution. The upcoming release will depend on two other libraries for encryption, and I would rather have new features enabled by default everywhere rather than have a complex code to handle the defaults.

QuentiumYT: Alright, I read that you will drop all packaging for Debian/Ubuntu & others, I guess you're on a rolling release or cutting edge Linux to require all of these packages? Does that mean the PPA will also be abandoned?
I will update the build from source according to this CMake switch

hluk: > I guess you're on a rolling release or cutting edge Linux to require all of these packages?

I use Fedora, which is not rolling, but moves somewhat faster than many distros (there is a bigger update every 6 months).

> Does that mean the PPA will also be abandoned?

Unfortunately, yes, I will drop PPA. It still uses very old and unsupported Qt 5 libraries.

I'm not sure if there is a way I can simplify the work to packagers - apart from mentioning any new cmake options and dependencies in the changelog.

The idea now is: if there are new build dependencies in the new release, to fail the builds until packagers include the new build deps. But at least then the new features will be available to users.

The alternative would be to disable all new features and rely on packagers to notice that the new build options. This may not happen until a user of the package complains.

QuentiumYT: Oh yes fedora, that explains the recent Qt versions and KF6 😄

This is sad but I understand that packaging is very time consuming. I wish debian building was simpler, I personally maintain a Qt app as well and spend countless hours learning all from scratch. Maybe let a warning on the PPA saying it's no longer maintained (hoping you plan to keep it for backward compatibility)

I really like having PPAs, maybe I can have a look at publishing for debian, without the latest features sadly but at least the latest version and Qt6 :)

Let me know if you're interested, I don't know how you built all the packages on all linux distros and versions tho. Did you setup a build matrix somewhere?

hluk: For Ubuntu, I generated and uploaded the necessary files to https://launchpad.net/~hluk/+archive/ubuntu/copyq. The process was:

1. Open an Ubuntu container
2. Clone source code repository and switch to the latest release tag
3. Run `utils/debian/create_source_packages.sh`
4. Run `dput ppa:hluk/copyq ../copyq_*.changes`
5. Check if the builds pass later

For other distros, including Debian, I upload source code tarball to openSUSE Build Service: https://build.opensuse.org/package/show/home:lukho:copyq/CopyQ

But still, all the packages use Qt 5.

QuentiumYT: Oh yes I see, seems very organized :) Yes Qt6 should be the preferred version yet!
I saw the openSuse build service tried to build with Qt6, why didn't it work?
Also, why not migrating the debian/control with build depends with qt6, I can help you with that. I don't know if the drop of all distros is definitive but maybe try to support less versions? On my project, I build only using jammy github runner so that LTS supports Debian and Ubuntu above works, I also use rpmbuild to directly convert to some rpm packages :)

hluk: > I saw the openSuse build service tried to build with Qt6, why didn't it work?

Ah, yes, that is the OBS project for Qt 6 builds (https://build.opensuse.org/package/show/home:lukho:copyq/CopyQ), which does not work. I would need to figure out the correct dependencies (package names) or disable the related features (probably somehow separately per each distro).

Here is the OBS project for Qt 5 builds, that the releases actually use: https://build.opensuse.org/package/show/home:lukho:copyq/CopyQ-Qt5

QuentiumYT: Oh yes alright, that was already Qt6 😄 I will have a look for building with it on launchpad using your scripts. Maybe we can automate a little bit more that to keep at least the launchpad PPA :)

QuentiumYT:
Alright, all good for the docs? Will have a look about Qt6 builds later :)

docs/build-source-code.rst
```diff
@@ -60,6 +58,11 @@ On **Ubuntu** you can install all build dependencies with:
       qt6-wayland-dev-tools \
       qtkeychain-qt6-dev
 
+On Ubuntu 25.04 and newer, you can also install `KF6` packages:
+    sudo apt install \
```
hluk, line 65: There should be an empty line before the code block, otherwise the formatting is broken.

```suggestion

    sudo apt install \
```

hluk:
LGTM, just one comment.

docs/build-source-code.rst
QuentiumYT: Added a codeblock to be more consistent :)

hluk: Oh, right, this is rst file, not md. Thanks! :)

hluk:
Looks good! Thanks! :+1:

PR #3402, opened 2026-01-29
Fix the JS comment syntax in a doc example

kurtmckee:
This resolves a warning emitted during doc builds [[recent example](https://app.readthedocs.org/projects/copyq/builds/31148357/#303436602--18)]:

```
WARNING: Lexing literal_block "let req = NetworkRequest();\n\n# allow redirects\n..."
as "js" resulted in an error at token: '#'. Retrying in relaxed mode.
```

Note that there are two additional doc warnings associated with INI syntax. I didn't see an obvious way to resolve those warnings.

Thanks for all of your work on CopyQ!

### DIFF
--- a/docs/scripting-api.rst
+++ b/docs/scripting-api.rst
@@ -2105,24 +2105,24 @@ Types
 
        let req = NetworkRequest();
 
-       # allow redirects
+       // allow redirects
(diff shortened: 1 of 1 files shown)


hluk:
Thanks for the patch!

PR #3406, opened 2026-01-31
Avoid overriding selection data in automatic commands

### LINKED ISSUE #3324: setData will modify the data of historical items
**Describe the bug**
When calling setData in an automated command, the history item also writes the same data.

**To Reproduce**
1. Fresh installation of CopyQ
2. Randomly copy a text once
3. Add and enable the following automated command
```ini
[Command]
Name=test setData
Command="
    copyq:
    setData('application/x-copyq-tags', 'test')"
Automatic=true
Icon=
```

4. Copy the text once again
5. Open the CopyQ main window and find that the data of both the current entry and the previous item have been modified

**Expected behavior**

Only the data of the current entry is modified

**Screenshots**

**Video**

https://github.com/user-attachments/assets/d307c948-64db-41b8-b5b6-205bf97c8883

- Application Version: 13.00
- OS: Windows 11
- Installation method: portable zip

**Additional context**

I want to store metadata (copy time, source url, window title, etc.) of entries in CopyQ. For this purpose, I created an automated command and called setData multiple times to write data with custom MIME types.

Later, I found that this command would overwrite the metadata of historical items. After troubleshooting, I identified the issue with setData, so I simplified the reproduction steps.

<details>
<summary>The original command</summary>

```ini
[Command]
Name=Record source data
Command="
    copyq:
    const MIME_TAGS = 'application/x-copyq-tags';

    const getClipStr = (mime, limit) => {
      const bytes = clipboard(mime);
      if (!bytes?.length) return null;
      // Only read the header
      return str(limit && bytes.length > limit ? bytes.mid(0, limit) : bytes);
    };

    const getDataStr = (mime) => {
      const val = data(mime);
      return val?.length ? str(val) : null;
    };

    const handlers = [
      {
        rawMime: 'CopyQ/copy-time',
        extract: () => dateString('yy-MM-dd hh:mm:ss'),
        tag: (time) => time,
      },
      {
        rawMime: 'CopyQ/source-url',
        extract: () => {
          return (
            // Remove trailing special characters
            getClipStr('text/x-moz-url-priv')?.slice(0, -1) ??
            getClipStr('HTML Format', 2000)
              ?.match(/^SourceURL:\\s*([^\\r\\n]+)/m)?.[1]
              ?.trim()
          );
        },
        tag: (urlStr) => new URL(urlStr).hostname,
      },
      {
        rawMime: 'CopyQ/window-title',
        extract: () => getDataStr(mimeWindowTitle),
        tag: (text) => text?.trim(),
      },
    ];

    {
      const newTags = [];

      for (const handler of handlers) {
        const content = handler.extract();
        if (!content) continue;

        if (handler.rawMime) {
          setData(handler.rawMime, content);
        }

        // Only add when the tag function exists and returns a valid value
        const tagVal = handler.tag?.(content);
        if (tagVal) {
          newTags.push(String(tagVal));
        }
      }

      if (newTags.length > 0) {
        // Prevent the original data from containing commas used as separator tags
        const safeTags = newTags.map((t) => t.replace(/,/g, '，'));
        setData(MIME_TAGS, safeTags.join(', '));
      }
    }"
Automatic=true
Enable=false
Icon=
```
</details>

hluk:
Fixes #3324

### DIFF
--- a/docs/scripting-api.rst
+++ b/docs/scripting-api.rst
@@ -804,9 +804,9 @@ unlike in GUI, where row numbers start from 1 by default.
 
 .. js:function:: setData(mimeType, data)
 
-   Modifies data for :js:func:`data` and new clipboard item.
+   Set data for automatic and display commands, or the current item selection.
 
-   Next automatic command will get updated data.
+   Next automatic command will receive the updated data.
 
    This is also the data used to create new item from clipboard.
 
@@ -833,7 +833,7 @@ unlike in GUI, where row numbers start from 1 by default.
 
 .. js:function:: removeData(mimeType)
 
-   Removes data for :js:func:`data` and new clipboard item.
+   Removes data for automatic and display commands, or the current item selection.
 
 .. js:function:: dataFormats()
 
--- a/src/gui/commandcompleterdocumentation.h
+++ b/src/gui/commandcompleterdocumentation.h
@@ -90,8 +90,8 @@ void addDocumentation(AddDocumentationCallback addDocumentation)
     addDocumentation("toUnicode", "toUnicode(ByteArray, encodingName) -> string", "Returns string for bytes with given encoding.");
     addDocumentation("fromUnicode", "fromUnicode(String, encodingName) -> `ByteArray`", "Returns encoded text.");
     addDocumentation("data", "data(mimeType) -> `ByteArray`", "Returns data for automatic commands or selected items.");
-    addDocumentation("setData", "setData(mimeType, data) -> bool", "Modifies data for `data()` and new clipboard item.");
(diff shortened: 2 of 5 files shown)


QuentiumYT:
Hey @hluk, just to warn you that you commit with 2 emails since you recently changed it. I don't know if that was intentional but I prefer to be sure

image

hluk:
> Hey @hluk, just to warn you that you commit with 2 emails since you recently changed it. I don't know if that was intentional but I prefer to be sure

Oh, thanks for the heads-up! I think it is because my other email address is set up on GitHub as Primary. I've just changed that.

PR #3461, opened 2026-03-05
Replace deprecated qt_add_translation function

QuentiumYT:
Hello, the qt_add_translation is deprecated (as per https://doc.qt.io/qt-6/qtlinguist-cmake-qt-add-translation.html)

Also while trying to package your app, I struggled with `set(copyq_DEFINITIONS ${${copyq_qt}Widgets_DEFINITIONS})`. It seems to look for extra widgets definitions but launchpad didn't work. Looks fine without it so I'm wondering if it's still needed.

Thanks for your reply :)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -61,7 +61,7 @@ set(copyq_COMPILE
 add_library(copyq-common OBJECT ${copyq_COMPILE})
 
 find_package(${copyq_qt} REQUIRED COMPONENTS LinguistTools)
-if("${QT_VERSION_MAJOR}" STREQUAL "5")
+if("${QT_DEFAULT_MAJOR_VERSION}" STREQUAL "5")
     qt5_add_translation(copyq_QM ${copyq_TRANSLATIONS})
 else()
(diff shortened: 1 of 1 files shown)


QuentiumYT:
Oh yes Qt5 doesn't support it, would you like me to add a fallback?

```cmake
if("${QT_VERSION_MAJOR}" STREQUAL "5")
  qt5_add_translation(copyq_QM ${copyq_TRANSLATIONS})
else()
  qt6_add_translations(copyq-common TS_FILES ${${PROJECT_NAME}_translations} QM_FILES_OUTPUT_VARIABLE copyq_QM)
endif()
```

hluk:
> Oh yes Qt5 doesn't support it, would you like me to add a fallback?

Yes, ideally, since the app still supports Qt 5. It is likely I can drop the support - it has been unmaintained for a while and I hope there are no maintained Linux distros without Qt 6.

QuentiumYT:
Here you go, I hope so but let's keep Qt5 support in case :)

hluk:
Thanks for the fix!

PR #3466, opened 2026-03-07
Speed up Windows and macOS builds

hluk:
Migrates Windows builds from Appveyor to much faster Windows runner on GitHub Actions.

Precompiles some Qt headers for Windows builds.

Updates dependencies on Windows: Qt 6.10.2, Snoretoast 0.9.1, KDE Frameworks 6.23

Optimizes some tests - avoids lower timeouts, fewer client process invocations.

Tries to improve performance macOS builds. Unfortunately the `macos-15-intel` image is very slow. Moving away from Homebrew reduces the build time from ~40 to ~20 minutes.


.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,186 @@
...
+          cache: true
+          set-env: true
+
+      - name: Install OpenSSL
+        shell: pwsh
+        run: |
+          choco install openssl --no-progress -y
+          # Chocolatey installs OpenSSL to 'C:\Program Files\OpenSSL' (not OpenSSL-Win64).
+          $sslDir = if (Test-Path 'C:/Program Files/OpenSSL-Win64') { 'C:/Program Files/OpenSSL-Win64' } else { 'C:/Program Files/OpenSSL' }
+          echo "OPENSSL_ROOT_DIR=$sslDir" >> $env:GITHUB_ENV
```
hluk: Is the conditional needed since the comment already states that OpenSSL is installed in specific directory?

```suggestion
          echo "OPENSSL_ROOT_DIR=C:/Program Files/OpenSSL" >> $env:GITHUB_ENV
```

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,186 @@
...
+          version: ${{ env.QT_VERSION }}
+          arch: win64_msvc2022_64
+          modules: qtimageformats qt5compat
+          cache: true
+          set-env: true
+
+      - name: Install OpenSSL
+        shell: pwsh
+        run: |
+          choco install openssl --no-progress -y
```
hluk: Can we cache this or install OpenSSL a different way? This takes over 1 minute - that seems a lot for this dependency.

PR #3482, opened 2026-03-16
Use newer metainfo instead of appdata

QuentiumYT:
As the title says :)

It's been quite some time the appdata format is deprecated, as well as some attributes in the XML file

Thanks


hluk:
Thank you!

PR #3501, opened 2026-03-25
Add One Dark theme

jetm:
## Problem

CopyQ ships several built-in color themes but does not include One Dark, a
popular dark color scheme used across many editors and terminals.

## Solution

Add a new `onedark.ini` theme file based on the One Dark color palette,
following the same structure as the existing `nord.ini` theme.

## Key changes

- Add `shared/themes/onedark.ini` with full One Dark color mapping
- Styled CSS for menu bar, tool bar, search bar, tabs, and item list
- Uses the standard One Dark palette: #282c34 background, #abb2bf foreground,
  #61afef blue selection, #e5c07b yellow search highlight, #e06c75 red accents

### DIFF
--- a/shared/themes/onedark.ini
+++ b/shared/themes/onedark.ini
@@ -1,0 +1,127 @@
+[General]
+css_template_items=items
+css_template_main_window=main_window
+css_template_menu=menu
+css_template_notification=notification
+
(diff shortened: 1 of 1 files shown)


hluk:
Thanks!

PR #3514, opened 2026-03-29
Add dependency versions and desktop environment info to version output

hluk:
Include KStatusNotifierItem, KGuiAddons, QCA and QtKeychain versions in the version output. Append full info() output including, on Linux, display-related environment variables (DISPLAY, WAYLAND_DISPLAY, XDG_CURRENT_DESKTOP, XDG_SESSION_DESKTOP, XDG_SESSION_TYPE).

Add QCA and QtKeychain to the About dialog library list.

Add Help > Copy Version and Info menu action for easy bug reporting.

Update bug report template and scripting API documentation.


src/gui/aboutdialog.cpp
```diff
@@ -127,6 +127,16 @@ QString AboutDialog::aboutPage(const Theme &theme)
             + helpLib("miniaudio",
                       "Copyright (c) David Reid",
                       "https://miniaud.io/")
+#endif
+#ifdef WITH_QCA_ENCRYPTION
+            + helpLib("QCA - Qt Cryptographic Architecture",
+                      "Copyright (c) Justin Karneges, Brad Hards, Ivan Romanov",
```
Chealer:
```suggestion
                      "© Justin Karneges, Brad Hards, Ivan Romanov",
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -27,11 +27,15 @@ Before continuing, make sure a similar request or a known issue does not exists
...
 **Version, OS and Environment**
 <!--
-Get details from `copyq version` command, or copy from the CopyQ main window:
-1. Open Action dialog (F5 shortcut).
-2. Enter command: `copyq: copy(version())`
-3. Click OK - the details should be copied to the system clipboard.
-4. Paste the details in the report.
+Get details using one of these methods:
+
+a. From the menu bar: Help > Copy Version and Info - the details are copied to the clipboard.
```
Chealer:
```suggestion
a. In versions above 14.0, from the main window’s Help menu, use “Copy Version and Info” to copy to the clipboard.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -27,11 +27,15 @@ Before continuing, make sure a similar request or a known issue does not exists
...
+b. From the command line: `copyq version`
```
Chealer:
```suggestion
b. From the command line: `copyq version` (but see [known issue on Windows](https://copyq.readthedocs.io/en/latest/known-issues.html#on-windows-copyq-does-not-print-anything-on-console))
```

Chealer (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
hluk: I'm not yet sure if this goes to 15.0.0 or 14.1.0.

Also, this is part of HTML comment (`<!-- ... -->`) - it's not Markdown itself.

.github/ISSUE_TEMPLATE/bug_report.md
hluk: Good point, thanks!

src/gui/aboutdialog.cpp
hluk: OK, let me apply it to all entries.

.github/ISSUE_TEMPLATE/bug_report.md
Chealer: Ah, sorry and thank you. I fixed my suggestion.

src/gui/menuitems.cpp
```diff
@@ -147,6 +147,10 @@ MenuItems menuItems()
                   QStringLiteral("help-contents"), IconCircleQuestion );
     addMenuItem( items, Actions::Help_ShowLog, QObject::tr("&Show Log"), QStringLiteral("show-log"), QObject::tr("F12"),
                   QStringLiteral("help-about"), IconCircleExclamation );
+    addMenuItem( items, Actions::Help_CopyVersionInfo,
+                 QObject::tr("&Copy Version and Info"),
```
Chealer: “Info“ is pretty vague. I suggest “Diagnostic information“ instead.
That being said, I am not convinced this warrants a menu item and default shortcut. I feel like the LibreOffice approach (in About), which also lets you see _what_ will be copied, is preferable.

[_License_](https://www.philippecloutier.com/Common+infrastructure+licensing#review)

.github/ISSUE_TEMPLATE/bug_report.md
hluk: How about the following?

```suggestion
a. In versions newer than 14.0.0, select “Copy Version and Info” from the Help menu to copy the details to your clipboard.
```

TBH, I never heard/seen "ulterior" being used anywhere but in "ulterior motives". 🙂

Chealer (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
Chealer: Good point, that is indeed better. My French origins are showing.
In fact, when rethinking about it, I thought it's not the date which matters, but the version number; I rectified my suggestion accordingly.
Thanks again

That being said, I notice that in option C we don't mention the relevant menu, but the shortcut. So I find A and C inconsistent in that regard. I feel navigating the menus is best (can be remembered for longer).

src/gui/menuitems.cpp
hluk: Yeah, you are right, this is a better idea - user would see what is being copied and can easily inspect the data in About dialog without pasting it anywhere first. I'm working on it.

docs/scripting-api.rst
```diff
@@ -724,7 +740,15 @@ unlike in GUI, where row numbers start from 1 by default.
 
 .. js:function:: info([pathName])
 
-   Returns paths and flags used by the application.
+   Returns diagnostic info: versions, features, platform, and environment
```
Chealer: Is the first “and” intentional?

src/common/diagnostics.cpp
```diff
@@ -0,0 +1,203 @@
...
+            "1"
+#else
+            "0"
+#endif
+        },
+        {"platform",
+#if defined(Q_OS_WIN)
+            "Windows"
+#elif defined(Q_OS_MAC)
+            "OS X"
```
Chealer: Nowadays "macOS”

src/gui/aboutdialog.cpp
```diff
@@ -76,6 +83,8 @@ QString AboutDialog::aboutPage(const Theme &theme)
         ".icon{font-family: \"" + iconFontFamily() + "\"}"
         ".help-icon{color:" + link + ";padding-left:1em;padding-right:1em}"
         ".library{font-size:12pt}"
+        "h3{font-size:14pt;margin-top:1.5em}"
+        "pre{font-size:9pt;white-space:pre-wrap;margin:0.5em 0}"
```
Chealer, line 105: The "pt" unit is not recommended for the screen (I suggest "px" instead): https://www.w3.org/Style/Examples/007/units.en.html

docs/scripting-api.rst
```diff
@@ -96,19 +96,35 @@ unlike in GUI, where row numbers start from 1 by default.
 
 .. js:function:: version()
 
-   Returns version string.
+   Returns versions, features and host system information.
 
    :returns: Version string.
```
Chealer: This is outdated… perhaps the line can be removed, since the function description already says it all?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -27,18 +27,20 @@ Before continuing, make sure a similar request or a known issue does not exists
...
+   - Desktop environment, window manager (if applicable)
+   - Installation method (for example Windows installer, portable zip, Flatpak, Homebrew, manual build)
+
+Paste the details in the report.
 -->
 
-- Application Version (for example 10.0.0)
-- OS (for example Windows 11, Ubuntu 24.04, Fedora 42, macOS Sonoma 14.4)
-- Desktop environment, window manager (if applicable)
-- Installation method (for example Windows installer, portable zip, Flatpak, Homebrew, manual build)
```
Chealer, line 40: I agree with removing all that _except_ the last line (installation method). Methods A to C do not provide it, so I guess it should be kept (unless perhaps it can be at some point automated, but that sounds non-trivial).

docs/scripting-api.rst
```diff
@@ -96,19 +96,35 @@ unlike in GUI, where row numbers start from 1 by default.
...
+       WAYLAND_DISPLAY: wayland-0
+       XDG_CURRENT_DESKTOP: KDE
+       XDG_SESSION_DESKTOP: KDE
+       XDG_SESSION_TYPE: wayland
+       has-audio: 1
+       has-encryption: 1
+       has-global-shortcuts: 1
+       has-keychain: 1
+       has-mouse-selection: 1
+       platform: Linux/X11
```
Chealer, line 127: I'm sorry if I'm just too outdated, but may I ask what exactly that platform is? Is it Wayland, X11 or both?

Chealer (commented):
Thank you again Lukas, this is a real improvement.

I tested the latest build and now got this much better output:

> Diagnostic Information
> CopyQ 14.0.0.17-g67aae77b
> Arch: x86_64-little_endian-llp64
> Audio: miniaudio 0.11.25
> Compiler: MSVC
> KNotifications: 6.23.0
> KStatusNotifierItem: 6.23.0
> OS: Windows 11 Version 25H2
> QCA: 2.3.10
> Qt: 6.10.2
> QtKeychain: 0.15.0
> has-audio: 1
> has-encryption: 1
> has-global-shortcuts: 1
> has-keychain: 1
> has-mouse-selection: 0
> platform: Windows

The only detail is `platform` is redundant with OS in this case. I am not sure I master current desktop Linux enough to recommend an alternative though.

Otherwise, I have a single concern: It took me a few seconds to spot the Diagnostic Information in the About dialog. There are several reasons for that:
1. I was expecting a button like LibreOffice’s.
2. It is out of the default viewport.
3. The viewport cuts in an empty place, so it's not obvious that the contents is incomplete, as the following screenshot shows.
4. The scrollbar is real easy to miss; there is a black bar at the top and at the bottom, so it only seems natural that there is one to the right.
5. It is not typical for About dialog to have scrollbars.

Image
#1 is specific to myself, on the other hand others will not _know_ that the information is somewhere, so I do worry _many_ will miss it.

#4 is definitely not specific to myself, but it is not the first time I notice that the scrollbar is very subtle in recent years. In my current Firefox, it is 20 times easier to miss than it would have been in Firefox 1… I am wondering if anyone has an opinion on that? Is that an issue with my Windows theme? The thing is I am using the default.

I wonder if I should personally do something about #4, but I would suggest the following:
1. Not set default dimensions on the dialog, so it uses (I assume) the space necessary to fit everything.
2. Use 2 columns. I hesitate on how to split though (perhaps library credits should be the right column?)

docs/scripting-api.rst
hluk: The X11 is very confusing, let me drop it. That just indicates that the app is build with X11 support.

I will probably drop this completely - more exact information is provided by the "OS" entry.

Chealer: > The X11 is very confusing, let me drop it. That just indicates that the app is build with X11 support.

Ah… I see, thank you very much. In that case, I agree with dropping (or presenting it differently).

.github/ISSUE_TEMPLATE/bug_report.md
hluk: It's not very important, if needed I can ask reporter.

src/gui/aboutdialog.cpp
hluk: Not sure if this is true for Qt widgets - "px" could mean physical pixel.

hluk:
This is how it looks like now: 
image

docs/scripting-api.rst
hluk: Updated.

docs/scripting-api.rst
```diff
@@ -96,19 +96,35 @@ unlike in GUI, where row numbers start from 1 by default.
...
 .. js:function:: version()
 
-   Returns version string.
+   Returns versions, features and host system information.
 
-   :returns: Version string.
+   :returns: Versions, features and host system information
    :rtype: string
 
    Example of the version string::
```
Chealer:
```suggestion
   Example return value:
```

src/gui/aboutdialog.cpp
```diff
@@ -33,13 +34,37 @@ QString helpLink(const QString &name, const QString &link, ushort icon)
 
 QString helpLib(const char *name, const QString &copyright, const char *url)
 {
-    return QString("<p><span class='library'>%1</span>"
+    return QString("<span class='library'>%1</span>"
                    "&nbsp;&nbsp;&nbsp;<br />"
-                   "<span class='copyright'>%2</span><br />"
-                   "%3</p>")
+                   "<span class='copyright'>\u00a9 %2</span><br />"
+                   "%3")
```
Chealer, line 40: The links are rather awkward when split on several lines🫤
I suggest integrating them in the title, but note that this will make it impossible to preview the target (without clicking), unlike in a browser.
```suggestion
    return QString("<a href="%3" class='library'>%1</a>"
"&nbsp;&nbsp;&nbsp;<br />"
                   "<span class='copyright'>\u00a9 %2</span><br />")
```

src/gui/aboutdialog.cpp
Chealer: At first look, you are probably right that px still means a physical pixel to QTextBrowser. Perhaps a reason to use QWebEngineView?🤔

Chealer (commented):
Thanks again, great look!

src/gui/aboutdialog.cpp
hluk: I would rather not. I don't want to pull in other dependencies and the About dialog seems OK to me using the simpler text widget.

src/gui/aboutdialog.cpp
hluk: I prefer seeing the link (without any tooltips). Should I revert to single column view?

docs/scripting-api.rst
hluk: Updated.

src/gui/aboutdialog.cpp
Chealer: Ah, right. Indeed, sorry and thanks

src/gui/aboutdialog.cpp
Chealer: OK
I think one of the following would be preferable:
1. Put the libraries on 2 columns instead of 3
2. Put libraries on their own column
3. Put libraries in a different dialog. JetBrains’s approach is interesting, in particular in CLion’s case (usage of tens of libraries). They don't show the link, but they do display the license.
4. Use tabs (as KeePassXC does).

src/gui/aboutdialog.cpp
```diff
@@ -99,39 +126,45 @@ QString AboutDialog::aboutPage(const Theme &theme)
...
-            + helpLib("Solarized",
-                      "Copyright (c) Ethan Schoonover", "https://ethanschoonover.com/solarized")
-
-        + "<p></p>"
-
-        "</body></html>";
+        + helpLibColumns(3, {
+            helpLib("Qt Toolkit",
+                    "The Qt Company Ltd. and other contributors",
+                    "https://www.qt.io/"),
```
Chealer: Pre-existing, but if I may, it was renamed to "Qt Framework".
```suggestion
            helpLib("Qt Framework",
                    "The Qt Company and other contributors",
                    "https://www.qt.io/development/qt-framework"),
```

Chealer (approved):

src/gui/aboutdialog.cpp
hluk: No problem, I appreciate the discussion and you help improve the quality. Thanks.

src/gui/aboutdialog.cpp
hluk: OK, 2 columns are fine.

Chealer:
Thanks again Lukas, I just tested the latest build and it's great. It would be even greater if the window was about ¼ larger by default.

PR #3533, opened 2026-04-04
Fix FAQ: clarify how to open context menu with keyboard

### LINKED ISSUE #3471: FAQ: Confusing question/answer “How to open the menu or context menu with only the keyboard?”
The FAQ contains [a confusing section](https://copyq.readthedocs.io/en/latest/faq.html#how-to-open-the-menu-or-context-menu-with-only-the-keyboard), which reads:

> ### How to open the menu or context menu with only the keyboard?
> 
> Use Alt+I to open the item menu or use the Menu key on your keyboard to open the context menu for selected items.

I struggled to make sense of both the question and its answer.

When CopyQ is in English, both <kbd>Alt</kbd>+<kbd>I</kbd> and <kbd>Menu</kbd> have very similar effects. The first opens the window’s _Item_ menu, while the latter opens the selected item’s contextual menu, but both of these menus have the same content.

When CopyQ is in French, <kbd>Alt</kbd>+<kbd>I</kbd> has no effect (<kbd>Alt</kbd>+<kbd>M</kbd> would be required instead).

This could be fixed by:
1. specifying which first menu the question refers to
2. capitalizing "item" and formatting it in italics
3. listing each locale's actual shortcut.

However, I wonder if that is really worth it; is that question really frequent? Mnemonics are ubiquitous, so it is hard to imagine power users unaware of them. The Menu key’s role is also an established convention, which I hope most CopyQ users know. Maybe there are a few who don't know that, and it would suffice to focus the question only on the context menu?

If the question is kept, it would help to indicate it is specific to the main window (perhaps by putting it in a section dedicated to the main window).

<details>
  <summary>License</summary>
This report (including all messages and attachments I add to it) is offered under the terms of CC0 1.0.
</details>

hluk:
Replace locale-specific Alt+I shortcut with locale-independent alternatives (Menu key, Shift+F10) and clarify the distinction between the context menu and menu bar access.

Fixes #3471

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -257,11 +257,11 @@ If pasting still doesn't work, it could be caused by either of these problems:
 - CopyQ fails to focus the target window correctly.
 - The format copied to the clipboard is not supported by the target application.
 
-How to open the menu or context menu with only the keyboard?
-------------------------------------------------------------
+How to open the context menu for selected items with only the keyboard?
+-----------------------------------------------------------------------
 
-Use ``Alt+I`` to open the item menu or use the ``Menu`` key on your keyboard
-to open the context menu for selected items.
+In the main CopyQ window, press the ``Menu`` key (or ``Shift+F10``) to open
+the context menu for selected items.
 
 .. _faq-hide-menu-bar:
 


docs/faq.rst
```diff
@@ -257,11 +257,14 @@ If pasting still doesn't work, it could be caused by either of these problems:
...
-------------------------------------------------------------
+How to open the context menu for selected items with only the keyboard?
+-----------------------------------------------------------------------
+
+In the main CopyQ window, press the ``Menu`` key (or ``Shift+F10``) to open
+the context menu for selected items.
 
-Use ``Alt+I`` to open the item menu or use the ``Menu`` key on your keyboard
-to open the context menu for selected items.
+To open the menu bar menus, press ``Alt`` and then the underlined letter in
```
Chealer: The pluralization is inconsistent.
```suggestion
To open a menu from the menu bar, press ``Alt`` and then the underlined letter in
```

This is pre-existing, and do not quote me on that since I never used macOS, but I do not think macOS has mnemonics. Therefore, either this should be marked as non-macOS… or removed? This is far from specific to CopyQ. In my opinion, this is the kind of stuff you should learn in high school.🤨

Chealer (approved):
Thank you, this fixes the reported issue(s).

docs/faq.rst
hluk: I'm removing mentioning the Alt key.

PR #3534, opened 2026-04-04
Fix missing console output on Windows

### LINKED ISSUE #349: No console output on Windows
Console executable `copyq.com` build with Qt 5 is unable to print anything on console.

~~Console output works with Qt 4 but when tests processes are started as `QProcess` then there doesn't seem to be anything printed to stdout or stderr.~~

~~If this is fixed and tests work as expected, test can be run using [AppVeyor](https://ci.appveyor.com/project/hluk/copyq) (#347).~~

**Edit:** The console app in Qt 4 seems to work fine (problem was release version didn't print debug info needed for passing keys in tests - fixed in commit e82f162186cb3561708ea5ba7eef380e4729ad25).

**Edit 2:** Tests run on AppVeyor but the whole process takes too much time and is killed before it can finish. Console output is still unavailable with Qt 5 (tested with `cmd` and `git-bash`).

### LINKED ISSUE #2937: --help not working on windows powershell, cmd or gitbash
**Describe the bug**
From CMD:
```
C:\Users\edgeb1>copyq --help
<no output>
```

From powershell
```
➜ copyq --help
<no output>
```

From Gitbash
```
edgeb1@RSCMW3512 MINGW64 ~
$ copyq --help
<no output>
```

**To Reproduce**
1. Run `copyq --help` from any available shell.

**Expected behavior**
Should display help options

image

image

that said, windows 11, copyq 9.1.0

image

- Application Version (for example 7.1.0)
9.1.0

- OS (for example Windows 11, Ubuntu 22.04, macOS Sonoma 14.4)
windows 11

- Desktop environment, window manager (if applicable)
windows 11

**Additional context**
I am able to run `copyq --add XXX` to add things to the clipboard.
I just get no output from any copyq shell commands

hluk:
Attach to the parent console when running as a client or console app so that output from commands like --help, --version, and scripting reaches the terminal. Piped output (e.g. copyq help | more) is preserved by checking handle types before reattaching.

Remove the corresponding known issue from documentation.

Fixes: https://github.com/hluk/CopyQ/issues/2937
Fixes: https://github.com/hluk/CopyQ/issues/349

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -132,6 +132,64 @@ jobs:
           APP_VERSION: ${{ steps.version.outputs.version }}
           BUILD_DIR: ${{ runner.workspace }}/build/copyq/Windows
 
+      - name: Test console wrapper (cmd)
+        working-directory: ${{ github.workspace }}/copyq-${{ steps.version.outputs.version }}
+        shell: cmd
+        run: |
+          @echo off
+
+          rem --- Test 1: stdout capture via redirect ---
+          copyq.com version > __out.txt 2>nul
+          if %ERRORLEVEL% neq 0 (echo FAIL: exit code %ERRORLEVEL% & exit /b 1)
+          findstr /C:"CopyQ" __out.txt >nul || (echo FAIL: no CopyQ in output & type __out.txt & exit /b 1)
+          del __out.txt
+
+          rem --- Test 2: piping works ---
+          copyq.com version 2>nul | findstr /C:"CopyQ" >nul
+          if %ERRORLEVEL% neq 0 (echo FAIL: pipe did not contain CopyQ & exit /b 1)
+
+          rem --- Test 3: exit code propagation for errors ---
+          copyq.com thisCommandDoesNotExist12345 >nul 2>nul
+          if %ERRORLEVEL% equ 0 (echo FAIL: expected non-zero exit code & exit /b 1)
+
+          rem --- Test 4: terminal still functional after wrapper exits ---
+          echo AFTER_WRAPPER | findstr /C:"AFTER_WRAPPER" >nul
+          if %ERRORLEVEL% neq 0 (echo FAIL: terminal corrupted & exit /b 1)
+
+          echo All cmd console wrapper tests passed.
+
+      - name: Test console wrapper (pwsh)
+        working-directory: ${{ github.workspace }}/copyq-${{ steps.version.outputs.version }}
+        shell: pwsh
+        run: |
+          # GitHub Actions prepends $ErrorActionPreference = 'Stop' to pwsh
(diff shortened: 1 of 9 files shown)


Chealer:
Wow, this seems complicated!

First, thank you very much for trying to fix this. This works, but with a big caveat: the prompt is printed before the output. At first, it looks like the command does not complete. This happens either with PowerShell in Windows Terminal, or with cmd.exe in Windows console. These are the only 2 combinations I tested🫤

This happens with `copyq.exe help` and `copyq.exe config`, the only 2 commands I tested.

Sometimes, all you need is to hit Enter, and you recover a sane console. But often, that will still leave output below, so it's rather messy.

I am afraid this goes beyond my CLI development and C++ skills, so I can't afford to advise. Still, if it's complicated to do better, I'm afraid that result is better than the current situation.

I checked quickly if I could find programs comparable to CopyQ. I don't have many programs, but am back quite emptyhanded. The only thing I found is [KeePassXC offers a `keepassxc-cli` command](https://keepassxc.org/docs/KeePassXC_UserGuide#_command_line_tool) for that. KeePassXC has Windows origins, but is built on Qt, so I don't see them as Windows-centric. I can't say I'm a fan of introducing a new command, but if they made that decision for all platforms, I suppose there was a strong reason.

In any case though, even if a proper fix is achieved, the known issue section should not be removed until the next version is released. In fact, I would advise to leave it for a few years. I worked for a couple ministries which only allowed to install approved packages, and their catalogue would only offer a Ditto version from a few years ago. I assume they did not offer CopyQ, but if they did, it certainly was not bleeding edge.

[_License_](https://www.philippecloutier.com/Common+infrastructure+licensing#review)

hluk:
@Chealer Thanks for testing! I could reproduce the problem with [Wine](https://www.winehq.org/) (`wine cmd`). I tried few other things, and it seems the only solution going back to using `copyq.com` executable for console output.

Honestly, I wouldn't be looking into this again without help of Claude Code. The solution seems crazy complex.

hluk:
With the new `copyq.com` included in the builds, I cannot really see any output in Wine (using `wineconsole` and `wine cmd`). It could be some Wine limitation - not sure if there is an automated way this can be verified.

hluk:
BTW, the latest build is here: https://github.com/hluk/CopyQ/actions/runs/23984984767?pr=3534

I'm not confident it is fixed though.

Chealer:
Thank you very much for all this effort Luklaude😁… but I’m afraid it's not just Wine.😣
It's back to square 1 according to the 3 combinations I tested:
1. PowerShell in Windows Terminal
2. cmd.exe in Windows console
3. Bash in mintty (Cygwin)

I thought about checking WinSCP, and it seems you were on the right track; unless calling `winscp.com`, it opens a child window. Which would still be better than nothing.

Surprisingly complex indeed🤨

PR #3535, opened 2026-04-05
Fix ECM version constraint and document optional dependencies

hluk:
Remove incorrect version constraint from ECM find_package call and document optional build dependencies (QCA, KNotifications, ECM, miniaudio) with their corresponding CMake disable flags.

Fixes #3518
Fixes #3525

### DIFF
--- a/docs/build-source-code.rst
+++ b/docs/build-source-code.rst
@@ -26,9 +26,25 @@ The build requires:
 - `CMake <https://cmake.org/download/>`__
 - `Qt <https://download.qt.io/archive/qt/>`__
 
-Optional:
+Optional (enabled by default, disable with the corresponding CMake option):
 
-- `miniaudio <https://miniaud.io/>`__ -- for built-in audio playback (``playSound``)
+- `QCA <https://invent.kde.org/libraries/qca>`__ (Qt Cryptographic
+  Architecture) -- for tab encryption (``-DWITH_QCA_ENCRYPTION=OFF`` to
+  disable)
+- `QtKeychain <https://github.com/frankosterfeld/qtkeychain>`__ -- for storing
+  encryption passwords in the system keychain (``-DWITH_KEYCHAIN=OFF`` to
+  disable)
+- `KNotifications <https://invent.kde.org/frameworks/knotifications>`__ -- for
+  native notifications on Linux/BSD (``-DWITH_NATIVE_NOTIFICATIONS=OFF`` to
+  disable)
+- `miniaudio <https://miniaud.io/>`__ -- for built-in audio playback
+  (``playSound``) (``-DWITH_AUDIO=OFF`` to disable)
+
+Conditional:
+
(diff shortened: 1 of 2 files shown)


docs/build-source-code.rst
```diff
@@ -26,9 +26,20 @@ The build requires:
...
+  disable)
+- `QtKeychain <https://github.com/frankosterfeld/qtkeychain>`__ -- for storing
+  encryption passwords in the system keychain (``-DWITH_KEYCHAIN=OFF`` to
+  disable)
+- `KNotifications <https://invent.kde.org/frameworks/knotifications>`__ and
+  `ECM <https://invent.kde.org/frameworks/extra-cmake-modules>`__ (Extra CMake
+  Modules) -- for native notifications on Linux/BSD
+  (``-DWITH_NATIVE_NOTIFICATIONS=OFF`` to disable)
+- `miniaudio <https://miniaud.io/>`__ -- for built-in audio playback
+  (``playSound``) (``-DWITH_AUDIO=OFF`` to disable)
```
Chealer:
```suggestion
- `KNotifications <https://invent.kde.org/frameworks/knotifications>`__ ―for native notifications on Linux/BSD
  (``-DWITH_NATIVE_NOTIFICATIONS=OFF`` to disable)
- `miniaudio <https://miniaud.io/>`__―for built-in audio playback
  (``playSound``) (``-DWITH_AUDIO=OFF`` to disable)

Conditional:
- `ECM <https://invent.kde.org/frameworks/extra-cmake-modules>`__ (Extra CMake
  Modules, required unless building for a platform other than X11 with ``-DWITH_NATIVE_NOTIFICATIONS=OFF``)
```

Chealer (approved):
Thank you very much, this is much better.

The suggested modification would be more accurate, since ECM is not quite an "option".

docs/build-source-code.rst
hluk: Thanks. I've clarified it. ECM is also needed for Wayland clipboard support.

docs/build-source-code.rst
```diff
@@ -26,9 +26,25 @@ The build requires:
...
+  native notifications on Linux/BSD (``-DWITH_NATIVE_NOTIFICATIONS=OFF`` to
+  disable)
+- `miniaudio <https://miniaud.io/>`__ -- for built-in audio playback
+  (``playSound``) (``-DWITH_AUDIO=OFF`` to disable)
+
+Conditional:
+
+- `ECM <https://invent.kde.org/frameworks/extra-cmake-modules>`__ (Extra CMake
+  Modules) -- for Wayland clipboard support (KGuiAddons) and native notifications
+  on Linux/BSD (can be disabled entirely on non-Linux platforms)
```
Chealer, line 47: Are native notifications really supposed to use ECM only on GNU? Currently, they use ECM on Windows (and presumably all platforms).

hluk: Ah, true. Fixed: https://github.com/hluk/CopyQ/commit/965d0d679d46004b5180fc9b832ea88362feea5a

Issue #3366, opened 2025-12-19
Select the second item in the clipboard history as default in the Tray menu


kpontussvensson:
Hi,

I'm trying to find a replacement for ClipX which doesn't seem to work very good with Win11. I've been testing a bunch of the alternatives and really like the simplicity of CopyQ when keybinding the tray-menu.

But is it possible to somehow have the default selected item in the clipboard history be the second one in the list? Since the first one is already available with normal paste (Ctrl + V), it's very nice and convenient with ClipX that the default selected item is the second one in the list. I find that it's a quite usual scenario that you alternate between two items in the clipboard history and it just really improves the productivity having it the way I described above.

Is this already available in the settings and I just missed it? Is it perhaps possible to implement it as an option?

Thanks!

Chealer:
Greetings,

> Is this already available in the settings and I just missed it? Is it perhaps possible to implement it as an option?

Surely not

This is an ITS. It is fine to suggest changes, but each ticket should first focus on documenting the issue. Please:
1. explain which issue you are asking to address
2. indicate at least 1 affected version.

I never tried ClipX, but if you are looking for a free software replacement, I recommend either Ditto (which I have used for 6 years) or CopyQ (which I am evaluating).

kpontussvensson:
> Greetings,
> 
> > Is this already available in the settings and I just missed it? Is it perhaps possible to implement it as an option?
> 
> Surely not
> 
> This is an ITS. It is fine to suggest changes, but each ticket should first focus on documenting the issue. Please:
> 
> 1. explain which issue you are asking to address
> 2. indicate at least 1 affected version.

I think I quite clearly documented the issue, if something is not clear then please point it out.

The version I'm testing out is the 13.0.0 version, but I'm sure it is the case of all previous versions as well since it seems like a weird feature to just remove.

> I never tried ClipX, but if you are looking for a free software replacement, I recommend either Ditto (which I have used for 6 years) or CopyQ (which I am evaluating).

Yeah, I think it's clear you didn't read my whole post.

Chealer:
> I think I quite clearly documented the issue, if something is not clear then please point it out.

There is nothing unclear in what you wrote; the problem is you did not describe any issue with CopyQ. To put it differently, you suggest a change, but you leave it to readers to figure out which benefit(s) it would provide.
 
> The version I'm testing out is the 13.0.0 version, but I'm sure it is the case of all previous versions as well since it seems like a weird feature to just remove.

[...]

PR #3552, opened 2026-04-12
feat: Default to second item in clipboard list

codekoriko:
## Summary
When opening the clipboard list, the first item (current clipboard content) is selected by default. This PR changes the default selection to the second item when available.

## Motivation
The first item usually mirrors the current clipboard, so selecting it by default is redundant. Selecting the second item enables a faster workflow: `<shortcut to open app> + Enter` pastes the previous clipboard entry.

That's what ClipX was doing and this behavior has been requested before: https://github.com/hluk/CopyQ/issues/3366

## Change
- Replace default `setCurrent(0)` with:
  `setCurrent(std::min(1, m.rowCount() - 1))`

### Before
- `<shortcut to open app> + Enter` → pastes the current system clipboard content
  (effectively the same as `Ctrl + V`)

### After
- `<shortcut to open app> + Enter` → pastes the previous clipboard entry
  (the item that was overwritten in the system clipboard)

### Selection logic
- If 2+ items → second item is selected
- If only 1 item → first item remains selected

## Notes
Minimal change, no impact on other selection logic.

### DIFF
--- a/src/gui/clipboardbrowser.cpp
+++ b/src/gui/clipboardbrowser.cpp
@@ -1725,7 +1725,7 @@ bool ClipboardBrowser::loadItems(const QByteArray &itemData)
 
     d.rowsInserted(QModelIndex(), 0, m.rowCount());
     if ( hasFocus() )
-        setCurrent(0);
+        setCurrent(std::min(1, m.rowCount() - 1));  // Select 2nd item if available, else 1st
     onItemCountChanged();
 
     return true;


hluk:
Thanks for the PR, but I don't think it is a good idea. This would not always work (e.g. when resetting the selection on Esc) and would be confusing. Why select second item in all tabs? Why select second item if the first one does not match clipboard?

I also often want to have the latest clipboard content item selected because I want to edit it or do other operations on it.

A better way could be to create Script command that overrides `onTabSelected` (and maybe `show`/`toggle` or other functions) and selected the second item if needed.

PR #3672, opened 2026-08-02
Harden sync, clipboard, and large-tab behavior

FreshSoftware4:
## Summary

This draft consolidates the CopyQ bugfix pass described in the supplied sync-bug exports and rebases it onto the current upstream `master` with normal ancestry.

It:

- makes ItemSync ownership, lazy materialization, atomic writes, URI imports, and filesystem reconciliation failure-aware
- prevents destructive item commands from deleting synchronized items whose backing data cannot be read
- hardens clipboard reads and retries, including Windows native format checks and WebP/GIF handling
- improves large-tab responsiveness with batched placeholder/layout work and sparse item-widget tracking
- fixes move-to-top-after-paste behavior, translation-independent shortcuts, window-state restoration, tag sizing, selected-tab v3 imports, and script-dialog sizing
- adds focused regression coverage for the affected sync, clipboard, image, import/export, shortcut, command, item, and script paths

## Why

The original failure mode combined network-backed synchronized items, transient or permanent read failures, and code paths that treated missing data as an empty item. That could allow later command/removal paths to discard an item even though its synchronized payload had never been materialized successfully. Related clipboard and large-tab behavior amplified the reliability and responsiveness problems documented in the reproduction exports.

During review, two additional holes were found and fixed:

1. command actions using `remove` still ignored typed selection-copy failures, including remove-only commands
2. trailing-row deletion could leave stale entries in the sparse item-widget index map

## Validation

- `git diff --check` passes
- the supplied Git bundle verifies as complete
- CMake was run with the repository required Debug/Ninja/test configuration under an initialized MSVC environment
- configuration reached dependency discovery but cannot complete on this machine because the Qt 6 development package (`Qt6WidgetsConfig.cmake`, Qt >= 6.2) is not installed
- runtime tests were therefore not run locally; the focused tests in this branch and Windows/SMB reproduction still need CI or a Qt-enabled environment

This is intentionally a draft while CI and maintainer review establish whether the broad bugfix pass should remain together or be split into smaller PRs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,10 @@
   its first frame as a generic image fallback for applications without WebP
   clipboard support (#3661).
 
+- Do not treat removal of CopyQ's private clipboard-owner marker as new
+  clipboard content, preventing pasted items from moving to the top when the
+  corresponding history option is disabled (#3646).
+
 - Avoids potential crash on fetching very large data in clipboard.
 
 - Wayland: Fixed portal shortcuts handling in non-default app sessions.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -71,6 +71,9 @@
   preventing some X11 window managers from cascading its position on each
   toggle (#3643).
 
+- Keep item tags within the attainable item width so right-aligned tags remain
+  visible when the main window is wider than the maximum item width (#3013).
+
 - Avoids potential crash on fetching very large data in clipboard.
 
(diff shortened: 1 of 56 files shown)


hluk:
> This is intentionally a draft while CI and maintainer review establish whether the broad bugfix pass should remain together or be split into smaller PRs.

It would be best if you can split this into small PRs. I probably won't have the time to review such a large changes.

PR #3685, opened 2026-08-23
Fix input method support with FakeVim enabled

### LINKED ISSUE #3676: XCompose key doesn't work (it seems)
If I'm not missing something, `XCompose` key stopped working in `copyq`.

**To Reproduce**

1. Create a new item or modify existing one...
2. Press `XCompose` key and try to enter some sequence of characters.
3. <the keys are entered as if `XCompose` key is not pressed>

- Arch Linux
- r6886.809038f6
- X11

hluk:
Instead of blocking all input method events, convert committed text
to FakeVim key input. This preserves proper undo grouping while
allowing compose sequences, virtual keyboards, and other IM-based
input to work.

Fixes: https://github.com/hluk/CopyQ/issues/3676

### DIFF
--- a/plugins/itemfakevim/itemfakevim.cpp
+++ b/plugins/itemfakevim/itemfakevim.cpp
@@ -14,6 +14,7 @@ using namespace FakeVim::Internal;
 #include <QMessageBox>
 #include <QMetaMethod>
 #include <QKeyEvent>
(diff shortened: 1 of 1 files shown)


aksr:
It works! (not sure about side-effects, though).

PR #3696, opened 2026-09-02
docs: add paste queue command example

wantosure:
## Summary
- add a command example for collecting clipboard items into a paste queue
- add a global shortcut command to paste queued items in first-copied order

## Testing
- git diff --check
- Not run: Sphinx docs build. The local Python environment has no Sphinx installed, and installing docs/requirements.txt failed because pip could not unpack wheels due to permission errors in the temporary directory.


hluk:
Thanks for sharing. :+1:

Issue #3583, opened 2026-04-29
Dialog Width Not determined as per previous releases


dbareis:
**Describe the bug**
I have a number of dialogs that display very thinly since I upgraded to copyq v14.0.0-gd935d521-v14.0.0, I'd skipped a few versions and think it was a v11 or so previously where the dialogs didn't have the problem.  I have added a "good" and "bad" example of the same basic dialog and the relevant difference seems to be the title length.  I assume in the past the dialog contents was taken into consideration.

I don't know if this new behaviour is expected or not, but I don't want to hardcode values where the contents width can be calculated UNLESS I have to as the dialogs will then always look fugly.

**To Reproduce**
See the dialog parameters in the two attached files.

[DialogWidthBad (short title).txt](https://github.com/user-attachments/files/27189616/DialogWidthBad.short.title.txt)
[DialogWidthGood.txt](https://github.com/user-attachments/files/27189617/DialogWidthGood.txt)

**Expected behavior**
I expect previous behaviours to presist to newer versions and more specifically the width of not just the dialog title to be taken into account. I also expected to be able to take a screenshot of the dialog....

**Screenshots**
For some reason I can't take a screenshot of the dialog window, it shoes up as an empty box. I had to take a picture of it from my phone.

<img width="600" height="422" alt="Image" src="https://github.com/user-attachments/assets/acfdd9e3-92f3-4e7c-b484-39d56b61a2ac" />

**Version, OS and Environment**
CopyQ Clipboard Manager 14.0.0-gd935d521-v14.0.0
Qt: 6.10.2
KNotifications: 6.23.0
Compiler: MSVC
Arch: x86_64-little_endian-llp64
OS: Windows 11 Version 25H2
Audio: miniaudio 0.11.25

Chealer:
Thank you for reporting, but it may help if you could provide actual reproduction steps or help bisecting. This could be a regression from version 12, in which issue #2947 was fixed.

Screenshots must be prevented by the _Hide from screenshots and recordings_ preference, in _General_.

PR #3700, opened 2026-09-09
Fix initial script dialog sizing from its completed layout

wyf-777:
Script dialogs without saved geometry can open at a narrow default width even when their labels need more space. In #3583, a short title leaves much of the form behind scrollbars. `inputDialog()` fills unspecified dimensions from the dialog's current size before it has adjusted to its contents, and adds the button box after sizing.

Create the button box and apply the stylesheet before calculating the default size with `adjustSize()`. Restore saved geometry and apply explicit `.width` / `.height` afterwards. This preserves user sizing and the scroll area's existing size limits for large forms.

Add integration tests for initial content fitting and for explicit dimensions followed by restoring a user-resized dialog. The test plugin exposes dialog geometry and a resize action to exercise these behaviors in the real application.

Related to #3583.

Validation on Windows 11 build 22631, MSVC 19.44, Qt 6.10.2, Debug:

- New content-fitting test fails on `81c5c21b60a3bc81c242eb0625a827e68470ad55`: horizontal scroll maximum is 124 instead of 0. It passes with the fix.
- `commandDialogFitsContents`, `commandDialogRestoreGeometry`, `commandDialog`, and `commandDialogCloseOnDisconnect` pass (6 passed including suite setup/cleanup, 0 failed).
- Running the form reconstructed from the issue attachment changes its initial size from 190×107 to 562×413 on this display. Large content remains scrollable.
- `git diff --check` passes.

Optional encryption, keychain, audio, native notifications and regular plugins were disabled for this local build. The test plugin was built and loaded. Full upstream CI, Linux/macOS and Qt 6.10.3 (currently used by Windows CI) have not been run. Existing saved small geometry is intentionally retained.

### DIFF
--- a/src/tests/tests_scripts.cpp
+++ b/src/tests/tests_scripts.cpp
@@ -6,6 +6,7 @@
 
 #include "common/mimetypes.h"
 #include "common/commandstatus.h"
+#include "common/display.h"
 
 #include <QElapsedTimer>
 #include <QRegularExpression>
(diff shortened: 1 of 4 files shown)


src/tests/tests_scripts.cpp
```diff
@@ -462,6 +463,53 @@ void CoreTests::commandDialog()
...
+        KEYS(clipboardBrowserId);
+        RUN_MULTIPLE(
+            [&]{ RUN(script, "DEFAULT\n"); },
+            [&]{
+                KEYS("focus::QLineEdit<dialog_saved size:QDialog");
+                const auto errors = m_test->runClient(Args() << sizeScript, expectedSize);
+                // Let the geometry guard unlock before simulating a user resize.
+                RUN("sleep(350)", "");
+                RUN("callPlugin('itemtests', 'resizeDialog', 'dialog_saved size', 420, 240)", "");
+                RUN("sleep(350)", "");
```
hluk: Better to avoid `sleep` calls if possible - these can cause transient test failures. Can you use `WAIT_ON_OUTPUT` macro instead?

hluk:
Thanks for the fix!

PR #3707, opened 2026-09-17
Windows: do not paste to a window the user cannot switch to

### LINKED ISSUE #2556: paste to current window not working on Windows
I'm on Windows 11, and "paste to current window" almost never works, not from the main window nor from the tray.
why almost never? it very very rarely pastes from main window, but never from tray (it would paste from main window 2-3 times, then would not work again, tried many times).
also, it seems there's a focus obtaining issue, since the last window is not focused (you can tell by the titlebar color)

**To Reproduce**
set in preferences both options for paste (from history (main window) or from tray).
either double click an item on the main window, or select an item from the tray.

**Expected behavior**
copyq should paste to the "current" (last) window

 - CopyQ version 7.1.0
 - Windows 11 22H2 22621.1848

**Additional context**
I saw https://github.com/hluk/CopyQ/issues/1395 and tried:
1. viewing the copyq log, there's nothing that stands out.
   sometimes this shows in the log:
   ```
   Warning [2023-11-26 22:20:50.965] <Server-24612>: [default] QtWarning: QSystemTrayIcon::setVisible: No Icon set
   Warning [2023-11-26 22:20:57.981] <Server-24612>: [qt.qpa.fonts] QtWarning: Unable to enumerate family ' "Font Awesome 6 Free 
   Solid (CopyQ)" '
   ```
2.  running `ClipboardOwnerDebug_1001` - no visible errors
3. shift+insert in the last window works

any help or debugging tips would be very much appreciated.

ntzb:
Fixes #2556

On Windows the window remembered as the paste target was often a titleless system window instead of the application the user came from, so "paste to current window" activated that window and sent the paste there. 
Nothing appeared in the log, which is why this has been hard to pin down.

Two causes:

`isAltTabWindow()` only rejected a few window classes. It accepted shell surfaces, cloaked store application windows and tool windows, which `EnumWindows()` reports as visible but which the user can never switch to. 
It now also rejects windows that are invisible, cloaked, tool windows, unable to take the focus, or without a title.

`getCurrentWindow()` returned the foreground window it had just rejected whenever the enumeration found no replacement, because the callback only assigns a window when it finds a suitable one. 
That let a single bad sample stick: raising it made it the foreground window, so every later sample found it again and the paste target never recovered. 
It now returns nothing, and the caller keeps the window remembered before.

Reproduced on Windows 11 with the target window being a titleless `WorkerW`-style window picked up while the tray menu was opening. 
Both the tray menu and main window paths paste correctly with this change.

`DwmGetWindowAttribute` is resolved at runtime so no new link dependency is needed.


src/platform/win/winplatform.cpp
```diff
@@ -235,16 +235,62 @@ HWND getLastVisibleActivePopUpOfWindow(HWND window)
...
+        return dwmapi
+            ? reinterpret_cast<DwmGetWindowAttributePtr>(
+                  GetProcAddress(dwmapi, "DwmGetWindowAttribute") )
+            : nullptr;
+    }();
+
+    if (!dwmGetWindowAttribute)
+        return false;
+
+    constexpr DWORD dwmwaCloaked = 14;
```
hluk: 14? Can you use `DWMWA_CLOAKED` instead? I ask because from documentation it seems the value is 13 - it is not stated explicitly, just counting the values (index) in [the enum](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute).

hluk (commented):
Thanks for the patch! LGTM, just minor comment.

src/platform/win/winplatform.cpp
ntzb: thanks for reviewing.
good catch on the magic number, switched to DWMWA_CLOAKED.
the enum has both DWMWA_CLOAK (13, the one you set to cloak a window) and DWMWA_CLOAKED (14, the one you read), and it starts at 1 rather than 0.
using the symbol from dwmapi.h avoids the question entirely.
including the header doesn't add a link dependency since the function is still resolved at runtime.

ntzb:
> Thanks for the patch! LGTM, just minor comment.

@hluk great :)
modified as required

